PR #30, opened 2016-01-19
Support for RUser 3.0.0

jtkendall:
This PR updates Helpy to use [RUser](http://github.com/jtkendall/ruser) 3.0.0 which removes all of it's previous dependencies (Hashie, RestClient, and MultiJson). It also removes the unnecessary `.location.` and `.name.` objects. This also moves to RandomUser API version 0.7 from 0.4.1.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -95,7 +95,7 @@ gem 'rails-timeago'
 
 gem 'faker'
 gem 'populator'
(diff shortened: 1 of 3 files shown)


scott:
Thanks for the update! I will add you to the list of contributors :)

PR #32, opened 2016-01-30
Updated german translation

zefanja:
Hi, 

I've updated the german translation. This is a formal translation. Do you like to have helpy in formal languages or informal ones. In german both is possible. It depends on what your audience is.


scott:
Hmmm I am not sure if it should be formal or not, but I think the answer is probably yes.  What do you think?

PR #37, opened 2016-02-16
Russian locale update

dmitryzuev:
I had updated untranslated strings, and also fix problem with one-two-other


scott:
Thanks for your updates!  I just pushed them onto demo.helpy.io so it should be working better now :)

PR #44, opened 2016-02-24
Various documentation improvements

olliebennett:
Appreciate that some of this overlaps with the resources at support.helpy.io but it seems documentation around developing should live in the repo.

If there are any specific areas of in-code documentation which need a once-over, I would be happy to take a look.


olliebennett:
Have made your suggested changes and rebased :)

PR #43, opened 2016-02-23
Setup Travis CI

olliebennett:
This PR enables [TravisCI](https://travis-ci.org/), to help ensure all subsequent PRs pass the test suite before merging.

It will require @scott to connect Travis to this repository, but that's trivial and free.

An example build (for my branch in this PR) is [here](https://travis-ci.org/olliebennett/helpy)... and it passes! :)

PR #46, opened 2016-02-24
Update Ruby to version 2.3.0

olliebennett:
Not sure if this is an option; are you hoping to support older Ruby versions? If not, the newest version has some interesting improvements around string freezing and null checking in particular.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,4 +9,9 @@ services:
 before_script:
   - psql -c 'create database helpy_test;' -U postgres
 
+script:
+  - bundle exec rake db:schema:load
+  - bundle exec rake db:test:prepare
+  - bundle exec rake
+
 cache: bundler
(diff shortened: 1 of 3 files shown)


scott:
I am kind of trying to stay in line with what the Digital Ocean rails image ships with, which I think is 2.2.1 right now, but i agree 2.3.0 is the preferred option.  What are some good rails/VPS hosts in the UK?

olliebennett:
Great :) I think in that case we should remove the `ruby` declaration from the `Gemfile` altogether, and (using Travis to verify compatibility for us), support at least a small range of Ruby versions so people can use the latest if they want to. I'll make that tweak here as soon as #43 is merged! :)

olliebennett:
Have investigated and read more about running builds against multiple Ruby versions on TravisCI. I think it's much simpler to just stick with the older version, not least because we'd need to fiddle about with the `Gemfile` and add complexity where it's not required.

You're right that `2.2.1` is the default on DigitalOcean so understand your argument for sticking to that.

There are a couple of other changes in this PR, so I'll rename it to be more appropriate!

PR #47, opened 2016-02-24
RuboCop setup

olliebennett:
[RuboCop](https://github.com/bbatsov/rubocop) has been an invaluable tool for our projects, and helps keep people from making silly mistakes. At minimum, it prevents syntax errors from slipping through, but can also be enabled more fully to control best practices when developing.

I've disabled most cops as there are lots of violations, and the list in `.rubocop.yml` will serve as the TODO list, which we can progress with by dealing with one cop at a time.

If we go ahead with TravisCI, we can integrate a RuboCop check there too, to highlight any problems early.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -65,11 +65,6 @@ Lint/HandleExceptions:
     - 'bin/rails'
     - 'bin/rake'
 
-# Offense count: 1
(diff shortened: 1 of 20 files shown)


scott:
I will take a look at this later this week when I have more time...

scott:
I like it.  I wish I knew about this before :)

PR #50, opened 2016-02-24
Replace LESS with SCSS

### LINKED ISSUE #45: Replace LESS with SCSS
This has a few benefits:

- We can remove `therubyracer` and therefore the `libv8` dependency, which are very clunky, and arguably overkill!
- SCSS has become the prevalent CSS extension (vs. LESS).
- Twitter Bootstrap is actually moving to SCSS for the next version and [Bootstrapv3 support](https://github.com/twbs/bootstrap-sass) is in place already.

My cursory glance indicates that the codebase is using only basic LESS constructs, which means it should be quick to switch to SCSS. Would you consider accepting a PR providing this refactor?

olliebennett:
This resolves #45.


olliebennett:
It appears that `admin.scss` is completely redundant; is that in-line with your thinking, or do the styles defined there apply somewhere I'm not aware of?

olliebennett:
I'm not aware of any stylistic regressions with this, although haven't fully explored the system. Seems odd that the `admin` stylesheet seemingly wasn't being used at all though, so perhaps I'm missing something.

olliebennett:
Hi @scott - I've simplified and fixed this now. Thanks for your help identifying the problem with `admin` stylesheets - as you can see, I've left the `admin` reference in `layouts/admin.html.erb` and removed it from being included on every page (in `application.scss`).

The `twitter-bootstrap-rails` gem is still required, and provides the helpers such as `add_breadcrumb` and `nav_bar`, but the CSS (via SCSS) and JavaScript components of Bootstrap are now provided by the `bootstrap-sass` gem.

olliebennett:
Great - a very keen eye you have! I've resolved the concerns you raised.

olliebennett:
(just rebased onto `master` too)

PR #51, opened 2016-02-24
Miscellaneous cleanup

olliebennett:


olliebennett:
Apart from the fact this wouldn't merge cleanly before, are there any problems? I've rebased so should be good to go if not! :)

scott:
I didn't have time to review all of them last night, let me review them a little more before pressing launch :)

olliebennett:
I've just rebased, so should be good now! :)

PR #58, opened 2016-02-27
SCSS Cleanup

olliebennett:
This builds upon work done in #50 - please merge that first to get a meaningful diff here.


olliebennett:
I was quite meticulous as I went through this, so am reasonably confident that I haven't deleted anything that should not have been deleted!

PR #74, opened 2016-03-04
Bundler Audit - Security Scanner (+ vulnerable gem update)

olliebennett:


olliebennett:
Hey @scott - this is good to merge; it adds a tool to check for vulnerable gems, and also updates Rails which is currently vulnerable. Thought it'd be a longer-term solution to implement a check rather than just updating as a one-off. For more info, see [bundler-audit](https://github.com/rubysec/bundler-audit).

PR #75, opened 2016-03-04
Resolving Brakeman vulnerabilities

olliebennett:


olliebennett:
This follows on from #72 and fixes the vulnerabilities that were identified by the [Brakeman](http://brakemanscanner.org/) gem.

Additionally, build will continue to fail until all vulnerabilities (both errors and warnings) are resolved.

app/controllers/admin_controller.rb
```diff
@@ -33,8 +33,9 @@ def articles
   end
 
   def update_order
-    object = params[:object].titleize.constantize
-    @obj = object.find(params[:obj_id])
+    # Safely identify the model we're updating the position of
+    klass = [Category, Doc].detect { |c| c.name.casecmp(params[:object]) == 0 }
```
olliebennett: Wanted to draw your attention to this and confirm that these are the only two models for which we can "update_order".

scott: That is correct as of that update.  I think there may be value in adding it to forums as well (just the forum display order- not topics) but for now lets stick with just Doc and Cat.

Issue #42, opened 2016-02-22
Two reply boxes appearing in admin under certain circumstances


scott:
When reply voting is enabled, and after an admin reply, two reply boxes appear.

![screen shot 2016-02-22 at 4 22 48 pm](https://cloud.githubusercontent.com/assets/14369/13236303/c0446400-d980-11e5-8732-749c4fa1bb86.png)

PR #80, opened 2016-03-07
Fix/refactor admin posts

### LINKED ISSUE #73: Cannot post from Q&A and Matrix forums
Not sure when this was introduced, but even when logged in, I cannot post messages from these views.

scott:
This PR refactors the `_post` partial by moving to two partials one for admin and one for front-end.  This removed a lot of icky spagetti switching code and made it easier to resolve some issues related to trying to share one partial between two concerns.

I also merged in the integration testing branch and extended it, so thats included in this PR as well.

This PR fixes #73, #42 and #20

### DIFF
--- a/app/views/layouts/_login.html.erb
+++ b/app/views/layouts/_login.html.erb
@@ -9,7 +9,7 @@
 
 <hr class="hidden-lg hidden-md"/>
 
-<div class="login-form" class="col-md-6">
+<div class="login-form col-md-6">
 <% else %>
 
 <div class="col-md-12 login-form" data-title="<%= t('devise.sessions.new.sign_in') %>">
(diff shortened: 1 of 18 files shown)


scott:
@olliebennett can you do a quick review before I pull this?

app/views/layouts/_login.html.erb
```diff
@@ -9,7 +9,7 @@
 
 <hr class="hidden-lg hidden-md"/>
 
-<div class="col-md-6">
+<div class="login-form" class="col-md-6">
```
olliebennett: Is this as reliable as `class="login-form col-md-6"`? No idea whether it's treated differently, but seems like the kind of thing to break in old browsers, for example. Can't find any info on what happens when an attribute is defined multiple times on an element like this.

olliebennett:
Nice one :) This looks good - glad to see some integration tests too!

app/views/layouts/_login.html.erb
scott: That looks like a mistake/typo good catch!

PR #85, opened 2016-03-10
[WIP/QA] Add commenting/discussions to docs

scott:
This PR adds optional commenting for chosen knowledgebase docs. @olliebennett code review please!

screen shot 2016-03-09 at 9 46 18 pm


lib/tasks/populate.rake
```diff
@@ -115,6 +115,39 @@ namespace :db do
...
+        doc_id: doc.id
+      )
+      post = topic.posts.create!(
+        body: Faker::Lorem.paragraphs(rand(1..2)).join('<br/><br/>'),
+        user_id: topic.user_id,
+        kind: 'first'
+      )
+
+      timeseed = rand(1..30)
+      Timecop.travel(Date.today-timeseed.days)
```
olliebennett: Doesn't impact much, but using `timeseed.days.ago` is more "Railsy", clear and robust, because it obeys the server's timezone. Not a problem here I suppose.

db/migrate/20160310044144_add_doc_id_to_topic.rb
```diff
@@ -0,0 +1,26 @@
+class AddDocIdToTopic < ActiveRecord::Migration
+
+  def up
+    add_column :topics, :doc_id, :integer, default: 0
+    add_column :docs, :topics_count, :integer, default: 0
+    add_column :docs, :allow_comments, :boolean, default: true
+
+    if Forum.where(name: 'Doc comments').first.nil?
```
olliebennett: Is there a reason the migration is adding a (seemingly random) Forum? I'm conscious that this will (obviously) run when anyone updates their Helpy instance; do they _need_ the 'Doc comments' forum? And what if they're not English? Could this be moved to `seeds.rb` or `populate.rake`?

app/views/topics/index.html.erb
```diff
@@ -47,8 +47,8 @@
 
 <div>
   <% if user_signed_in? %>
-    <span class="hidden-xs pull-right"><%= link_to t(:open_new_discussion, default: "Open Discussion"), new_topic_path, class: 'btn btn-primary' %></span>
+    <span class="hidden-xs pull-right"><%= link_to t(:start_discussion, default: "Start Discussion"), new_topic_path, class: 'btn btn-primary' %></span>
```
olliebennett: We've renamed `:open_new_discussion` to `:start_discussion` but have the i18n file keys been changed too?

olliebennett: And on a related note, I feel that scattering the `default: 'Start Discussion'` all over the place is not DRY, and could be avoided if we _required_ that key to be in place? The app config should take care of using the `EN` translation as default if not present in the current language, and - to avoid the risk of "no translation found" blank strings - we could use Rails' `config.action_view.raise_on_missing_translations = true` setting to crash and bring these to our attention during testing.

olliebennett:
Looks good, @scott ! Have a glance over my comments but :+1: otherwise! :)

lib/tasks/populate.rake
scott: Thats true, more readable too.

db/migrate/20160310044144_add_doc_id_to_topic.rb
scott: I think there needs to be a forum for comments on the documents, which is what this is.  I did add it to the 'seeds.rb` but for people upgrading, we have to ensure the forum gets added.  Any other ideas on where to put it?

app/views/topics/index.html.erb
scott: Its actually not renamed, those are two different keys with slightly different semantics.  One of the things I tried to do here was not introduce new keys, which would then be unavailable in the other languages (or we would have to go chase down translations for.)  

I actually included the default text everywhere not to avoid crashing or exceptions- but as a matter of style- I think it makes the html more readable for people who may not be familiar with the project...

db/migrate/20160310044144_add_doc_id_to_topic.rb
olliebennett: Okay, that makes sense I suppose. Does the forum contain _all_ comments for _all_ documents or something? I don't fully understand the use-case for this, but agree that if we need this it'd seem there's no better place for it than in a migration.

app/views/topics/index.html.erb
olliebennett: Great, okay - understood. Agree it's good to avoid adding new keys if at all possible - I should have looked more closely before commenting, sorry.

I now see your reasoning for specifying the "Start Discussion" text everywhere too, and it makes sense. I probably still wouldn't do it, because - especially in this case - the key matches the text, so it's not a big leap to see what it'll say (and it'll need updating everywhere if we change the text, or else it'd then be misleading), but happy with it whichever way you like :)

olliebennett:
:+1:

db/migrate/20160310044144_add_doc_id_to_topic.rb
scott: Well, since the doc discussions are taking place on the actual doc show view, I don't want them showing up in the public community area, hence the need for a private forum or other way to classify them and filter them out.

PR #86, opened 2016-03-10
Minor css improvements

christ0ph3r:


christ0ph3r:
Hi cool project.  I am thinking of using this.  I was going through css and noticed that some of the rules had single quotes around them.  I haven't seen that before but it seemed like a typo.

Also I have no idea if trix-editor is a class or and ID but I am assuming its one of them?  Had a hard time figuring out which one or if its even in use still.

scott:
Hey Chris,  Thanks for submitting this!!  @olliebennett is currently working on some refactoring of the CSS and linting it, so I will wait on your PR until he gives the green light.  Trix-editor is the new wysiwyg editor from 37signals.

olliebennett:
Well noticed on the strange quote issue you fixed; agree that's a bug that should be fixed, although not sure what styling it would affect.

The `trix-editor` bit is correct without the class indicator I think (i.e. `trix-editor` instead of `.trix-editor`). As @scott says, it's the text editor, and is being inserted automatically by the `trix` gem, using code like `<%= f.trix_editor ...` in `app/views/docs/_form.html.erb`, so don't know a better way to style it. We shouldn't change this I don't think.

If you're interested, my related PR is #90 where I'm restructuring some of the CSS into proper SCSS format, and implementing some stricter automatic linting rules so that code style can be maintained better, with the help of scss-lint (which is awesome)!

scott:
@olliebennett This has conflicts after your PR was merged in, do you think your PR hit these couple things @christ0ph3r changed?

olliebennett:
Hey, yes it does look like they've conflicted - sorry! I've quickly resolved the conflict and created a separate PR as #93 - the commit is still attributed to @christ0ph3r but happy to close that if you'd prefer to fix the merge conflict yourself here Chris? Didn't want to interfere but wanted to fix the conflict I caused ;)

In that PR I also reverted the `.trix-editor` change, because that breaks the styling applied to the [New Content](http://localhost:3000/admin/docs/new?lang=en&locale=en) page.

christ0ph3r:
Hi guys.  Thank you very much.  I haven't gotten to set this up yet and have been busy this week. I am going to set this up in the next few days.  I just realized removing the single quotes actually looks like it will show changes http://codepen.io/anon/pen/QNKKpX .  I probably submitted a PR way too early on this one. Sorry about that.  If you do happen to like it with the new changes then feel free to merge 93.

PR #90, opened 2016-03-10
SCSS Lint Improvements

olliebennett:
I've merged several selectors and converted some CSS into SCSS format, without deliberately changing any behaviour or actual styles.

### DIFF
--- a/.scss-lint.yml
+++ b/.scss-lint.yml
@@ -14,3 +14,54 @@ linters:
   # Forbid empty blocks
   EmptyRule:
     severity: error
+
+# TODO: Enable the following
+
+  ColorKeyword:
+    enabled: false
+  ColorVariable:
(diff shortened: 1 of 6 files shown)


scott:
Do you feel like this is ready to merge?  I will pull it to local and review it to see if I see anything wrong first. I won't get to it until later today...

olliebennett:
Yeah - I have done some looking around and nothing looks out of place. The changes do not affect the actual rules themselves, but due to moving things around a bit, the precedence of different rules may have changed slightly, which is why it's not 100% guaranteed to behave identically.

Appreciate you having a look around before merging too :)

scott:
I checked it out side by side and checked the views I thought could be affected by the changes and I couldn't spot any problems, so merging!

Issue #69, opened 2016-03-03
Creating a new ticket as an admin and leaving the email field blank, breaks the open tickets view


rschaerer:
Creating as an admin a new ticket on behalf of a customer via the admin interface (`/admin/tickets/new`) and leaving the `Email` field blanks breaks the view of "new tickets") (`/admin/tickets?locale=de&status=new`) because the database field `user_id` in the tables `topics` and `posts` is `NULL`.
# Proposed solutions to fix this
- make the email field mandatory
- assign a `no e-mail provided` user to tickets witout an e-mail entered

Peronally, I'd prefer the second option, because we have sometimes customers without an e-mail and just a phone numer. If this bug get's fixed by implementing the first suggestion, then we would have to enter ad dummy e-mail address by ourself to record such tickets. this would lead in missinformation about users, because every ticket would be assigned to the same dummy email address user.
# More context, if needed:
- When a non logged in user tries to post a new ticket on `/de/topics/new` without entering data in the email field, the form returns an error and everything works as expected.
- To fix tickets without an email entered on the database level I had to assign the entry in the `topics` table with an `user_id`other than NULL. After doing this, I could view the new tickets view again, but the post isn't viewable too when i try to click it. To fix this, I had to assign once more an `user_id` to the entry in the `posts` table of my PostgreSQL database.

scott:
Thanks for this, we'll take a look at see what we can do to fix it :)

olliebennett:
This is certainly an interesting problem.

I'd propose that a short-term fix (to avoid actual breakage of the system) would be to require the email input by the admin when creating the ticket.

As for the longer-term solution, I agree it'd be nice to avoid admins having to enter `dummy@example.com`.

My thought on some (non-ideal) options:

- One single "no-email" user. Not sufficient because - to use the example above - the user opening the ticket may have other data (such as a phone number) which would want to be tracked on a per-user basis and not on the single "no-email" user.

- One new "no-email" user automatically created per "email-less ticket". This would let us store the phone number etc. on a per-user basis. This would be a tricky one to implement, as `devise` requires the presence of `email` to allow for all the registration / validation / password reset / confirmation etc. that comes as standard. If we just generated something like `user-<random-numbers>@example.com`, that could work. It might get ugly as we'd have to think how to display these sorts of email, and/or at least indicate that they were "fake" somehow.

[...]

PR #91, opened 2016-03-10
Misc. Enhancements

scott:
This PR adds two things- 

1. It uses javascript to disable submit buttons for creating and replying to topics.  This will resolve a couple issues- like #69.  It prevents the UI from going wonky by making the email mandatory.  The downside is that this is only fixed on the client side, the underlying problem still exists.

2. Auto-link URLS that exist in topics.  Adds a style to make links more visible in text content areas.


olliebennett:
Looks like you've introduced **Bootstrap** via Bower; could we remove that from the repo, and any idea why it's being added? I don't see any mention of dependencies in Helpy's `bower.json`, likewise `autolink-js` does not depend on Bootstrap.

scott:
I am not sure why it got added, but it did install with autolink-js.  I am pretty sure it must be a dependency of another package and Bower was just trying to meet all dependencies.  It also reinstalled font-awesome as a bower package, which is not needed.  I will prune both of them out of the PR.

olliebennett:
Nice one. If we use `bower install my-package --save` then it'll update the `bower.json` file and help us keep track of the versions etc. that we're depending on, which makes it slightly easier for others to perform updates of dependencies etc.

PR #97, opened 2016-03-16
fix locale on failure auth

### LINKED ISSUE #41: Devise not redirecting to correct locale
If an unlogged in user creates a new ticket in a non default language, they are redirected to the sign in screen in the default language.

rusikf:
Fix #41  + `rake rails:update:bin`

### DIFF
--- a/config/initializers/devise.rb
+++ b/config/initializers/devise.rb
@@ -1,5 +1,7 @@
 # Use this hook to configure devise mailer, warden hooks and so forth.
 # Many of these configuration options can be set straight in your model.
+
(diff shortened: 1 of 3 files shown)


scott:
Hi @rusikf Thanks for fixing this issue, I will merge it in once I have a chance to test it out. What is the change with rails:update:bin removing spring?

rusikf:
@scott It looks,that it was a [strange rvm error]( https://github.com/rails/rails/issues/14115) , but after ```bundle exec spring binstub --all spring ``` spring added, so I force updated my commit to remove bin fixes

PR #102, opened 2016-03-20
fix reply in posts

### LINKED ISSUE #79: If reply voting is enabled, reply form does not work
In a forum thread where reply voting is enabled, the inline reply forum does not submit unless you reload the page.  I narrowed it down to a turbo links issue, but didn't have time to dig further yet.

rusikf:
fix #79 
```$('.profile').initial();``` - reinitialize avatar
```<div id="posts">``` - valid markup, there is now rows in table, so posts are not wrapped in table#posts( see page source before this commit in chrome), so replace table with div.


scott:
Thanks for the patch!  I will take a look at it :)

PR #103, opened 2016-03-22
add regexp for username validation

### LINKED ISSUE #100: User name should not allow numbers.  Initial.js is not handling numbers in names and freezes.
A username should not allow numbers.  Currently numbers are permitted, and this is messing up initial.js, so may need to patch this also.

rusikf:
fix #100 
With this pr, following usernames are valid:
Abraham Lincoln
abraham lincoln
abraham
Any other combinations with numbers, or etc symbols are not allowed.
@scott What yout think about it?


scott:
@rusikf This is pretty much what I was thinking.  Could you add a unit test or two to describe the validation?  Also, if you are interested shoot me your email address and I'll invite you to our contributors slack, would love to chat a bit.

olliebennett:
What happens if I'm called `Björk Guðmundsdóttir`? I think it's important we all read [this](https://www.w3.org/International/questions/qa-personal-names) before getting carried away with name validation. I feel it's especially important given the effort we're going to to support alternative localisations in Helpy.

Was there a specific reason for this validation?

scott:
So I thought of a common naming occurance (at least here in the US that would not pass this- a hyphenated last name.  Fairly common to have last names like "Day-Smith" or any other concatenation.  @rusikf can you modify to support that and we'll try to come up with a bunch of cases to test against.

rusikf:
Does we need only "-" or something else ? As noticed @olliebennett in slack , it is a initial.js bug , but I can only fix initial.js and reject changes in this pr , what you think?

scott:
I wrote a unit test for the `User.name`, using the W3C article @olliebennett shared as a base, then added a couple additional cases of my own.  I still think its a great idea to finish this validation @rusikf to block number in names.  Fixing initial.js is a separate issue.

rusikf:
@scott, ok I added ability for hyphens both for last and first names ( first names can be at the end of sentence two)

scott:
Looks like its still failing the unit test I added.  Here are the test cases if you didn't see them:

   ```
 names = [
      "Björk Guðmundsdóttir",
      "Isa bin Osman",
      "毛泽东",
      "María-Jose Carreño Quiñones",
      "Борис Николаевич Ельцин",
      "Наина Иосифовна Ельцина",
      "John Q. Public",
      "V. S. Achuthanandan",
      "Nguyễn Tấn Dũng",
      "Steve Johns-Smith",
      "Scott",
      "JJ Adams",
      "T O'Shea"
    ]
```

rusikf:
@scott , ok, I forgot about new commits in master branch, fix it.

PR #106, opened 2016-03-25
[WIP] Basic Admin Statistics Page 

olliebennett:


app/views/admin/stats.html.erb
```diff
@@ -0,0 +1,51 @@
...
+
+<%
+
+# See: http://stackoverflow.com/a/1341318/1323144
+def mean(array)
+  array.inject{ |sum, el| sum + el }.to_f / array.size
+end
+
+# See: http://stackoverflow.com/a/14859546/1323144
+def median(array)
```
olliebennett: Hopefully a temporarily solution; would be nice to do it within the database instead.

scott:
Nice start!  I would like to get to a point where we have this broken out for each admin/agent, as well as the support team as a whole.  I gotta think there is a cleaner way to get mean and median, if not in the DB, at least as some kind of core ruby extension/gem.

app/views/admin/stats.html.erb
```diff
@@ -7,73 +7,88 @@ def mean(array)
...
+            <td><%= posts_count.zero? ? '-' : posts_count %></td>
 
-            <% delays = Topic.undeleted.where(assigned_user: t.assigned_user).where('posts_count > 1').map { |t| t.posts.second.created_at - t.created_at } %>
-            <td><%= distance_of_time_in_words median(delays) %></td>
+            <% delays = @responded_topics.where(assigned_user: t.assigned_user).map { |t| t.posts.second.created_at - t.created_at } %>
+            <td>
+              <% if delays.empty? %>
+                N/A
+              <% else %>
+                <%= distance_of_time_in_words median(delays) %>
```
scott: I got a nomethod error on the call to median()

```
NoMethodError - undefined method `median' for #<#<Class:0x007f935da0b980>:0x007f935f4bf768>:
  app/views/admin/stats.html.erb:91:in `block in _app_views_admin_stats_html_erb___3228140751521401155_70135467900280'
  activerecord (4.2.6) lib/active_record/relation/delegation.rb:46:in `each'
  app/views/admin/stats.html.erb:56:in `_app_views_admin_stats_html_erb___3228140751521401155_70135467900280'
```

PR #108, opened 2016-03-25
fix textarea width + locale

rusikf:
I think it will be nice to set full width for textarea full width for textarea + fix locale
Related to #107

### DIFF
--- a/app/views/topics/new.html.erb
+++ b/app/views/topics/new.html.erb
@@ -25,7 +25,7 @@
   <% end %>
 
   <%= f.text_field :name, :size => 40, placeholder: I18n.t(:subject), label: 'Subject', class: 'disable-empty' -%>
-  <%= text_area :post, :body, :rows => 8, :cols => 60, placeholder: I18n.t(:how_can_we_help), label: 'Message', class: 'disable-empty' -%>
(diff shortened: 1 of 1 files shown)


app/views/topics/new.html.erb
```diff
@@ -35,7 +35,8 @@
   <!--<label>Tags:</label><br/>-->
   <%#= text_field_tag :tags %>
 
-  <%= f.submit t(:submit_start_discussion, default: 'Start Discussion'), class: 'btn btn-warning disableable', disabled: 'disabled' %>
+  <%= f.submit t(:submit_start_discussion, default: t('submit_start_discussion')), class: 'btn btn-warning disableable', disabled: 'disabled' %>
```
scott: Nice catch :)

app/views/topics/new.html.erb
```diff
@@ -35,7 +35,8 @@
...
   </div>
```
scott: The default provides backup text in case the i18n files don't have the translation.  Putting the translation key there again would probably cause an error if the string was not translated?  I think this needs to stay the same as it was.

scott:
Thanks! I agree about the text area.  There are probably others that need this class as well.  Also were you able to test out the widget on any other sites?

rusikf:
@scott I reverted default text, and force pushed it again, no I don't tested it yet

scott:
Awesome, thanks @rusikf

PR #110, opened 2016-03-26
Add griddler-sparkpost for SparkPost integration.

PurpleMadness:
Sparkpost is a populair mailservice after Mandrill has announced their Mailchimp merge, so it would make sense to add it by default. Tested it and works on my installation

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -81,6 +81,7 @@ gem 'griddler-sendgrid'
 gem 'griddler-mailgun'
 gem 'griddler-postmark'
 gem 'griddler-mailin'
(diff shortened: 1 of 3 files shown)


scott:
@PurpleMadness Thanks for sending this, and thanks for adding the lock file too :)

scott:
Hi @pdewit - Can you tell me how you configured this in Sparkpost?  Which webhook checkboxes did you use?  I am struggling to get it working now for some reason...

pdewit:
I'm not sure, I don't think I did anything special to get it working. From what I can remember it was pretty straightforward.

PR #113, opened 2016-03-30
[QA] Settings in Admin

scott:
Needs some tests.  Want to add languages too.  Code review.


app/views/topics/index.rss.builder
```diff
@@ -2,7 +2,7 @@
 xml.instruct! :xml, :version => "1.0"
 xml.rss :version => "2.0" do
   xml.channel do
-    xml.title "#{Settings.site_name} #{@forum.name} Feed"
+    xml.title "#{AppSettings['settings.site_name']} #{@forum.name} Feed"
```
rusikf: ```AppSettings['settings``` sounds strange for me, @olliebennett what you think ?

app/controllers/admin_controller.rb
```diff
@@ -402,6 +402,36 @@ def user_search
...
+  def update_settings
+
+    @settings = AppSettings.get_all
+
+    # iterate through
+    @settings.each do |setting|
+#      logger.info("Setting: #{setting[0]}")
+#      logger.info("Value: #{params[setting[0]]}")
+
+      AppSettings[setting[0]] = params[setting[0].to_sym]
```
rusikf: I am think ```params``` hash works nice with strings and symbols as keys, why to use `to_sym` here ?

app/views/topics/index.rss.builder
scott: Unfortunately this is the API that is available from the rails settings cached gem.  It is not the prettiest way to express a namespaced variable.  About the only ooption would be to use a different name.

PR #131, opened 2016-04-26
Vote counter refactor

tapalilov:
Hey. I`ve digged in source and found piece of code that i can improve it.
I will be glad if you like it.

Best regards,
Georgi.


scott:
Nice optimization @tapalilov Thanks for the PR!

PR #138, opened 2016-05-09
thumbnails displays for pdf

andela-ysanni:


app/helpers/application_helper.rb
```diff
@@ -80,4 +80,8 @@ def css_overrides
     styles += "</style>"
     styles.html_safe
   end
+
+  def get_path(screenshot)
+    screenshot.format == "pdf" ? "#{screenshot.public_id}.png" : screenshot.path
```
scott: Did you mean `.png`?

scott:
Nice work!  I have one question above.  Also, if you are up for it, a nice refactor would be to replace this code block with a helper to dry up those partials: 
``` 
 			<div id="post-images">
  				<ul class="list-inline">
  					<% comment.screenshots.each do |screenshot| %>
						<li><%= link_to cl_image_tag(get_path(screenshot), { size: '125x125', crop: :fit }), "http://res.cloudinary.com/#{Cloudinary.config.cloud_name}/image/upload/#{screenshot.path}", class: 'screenshot-link' %></li>
  					<% end %>
  				</ul>
  			</div>
```

app/helpers/application_helper.rb
andela-ysanni: Yeah but it can also be .jpg
you can check this link for more details if what I did is not clear enough: http://cloudinary.com/blog/uploading_converting_and_generating_thumbnails_for_pdf_documents

app/views/admin/_post.html.erb
```diff
@@ -22,14 +22,8 @@
...
-			<div id="post-images">
-				<ul class="list-inline">
-					<% post.screenshots.each do |screenshot| %>
-						<li><%= link_to cl_image_tag(get_path(screenshot), { size: '125x125', crop: :fit }), "http://res.cloudinary.com/#{Cloudinary.config.cloud_name}/image/upload/#{screenshot.path}", class: 'screenshot-link' %></li>
-					<% end %>
-				</ul>
-			</div>
-			<% end %>
+              <%= render :partial => 'posts/thumbnail', locals: { :model_name => post } %>
+            <% end %>
```
scott: Could we take this one step further with
```<%= render :partial => 'posts/thumbnail', locals: { :model_name => post } unless Cloudinary.config.cloud_name.nil? %>```

scott:
Thanks Yetty this is awesome!

PR #148, opened 2016-05-16
Mailman support - Update PR after changes in Master

jprberlin:
Use latest master

Had to create a new PR cause i has some hassles with Git


scott:
Hi Jan, thanks for refactoring.  One request for the UI:

screen shot 2016-05-17 at 9 50 53 am

app/views/admin/settings/index.html.erb
```diff
@@ -86,8 +86,19 @@
...
+          <%= f.text_field 'email.pop3_username', value: AppSettings['email.pop3_username'], label: "POP3 Login" %>
+          <%= f.password_field 'email.pop3_password', value: AppSettings['email.pop3_password'], label: "POP3 Password" %>
+        </div>
+        <div class="imap-settings <%= "hidden" if "#{AppSettings['email.mail_service']}" != 'imap' %>">
+          <p>IMAP will require run 'rake helpy:mailman'. Also changes will require a restart of the rake task</p>
+          <%= f.text_field 'email.imap_server', value: AppSettings['email.imap_server'], label: "IMAP Host" %>
+          <%= f.text_field 'email.imap_username', value: AppSettings['email.imap_username'], label: "IMAP Login" %>
+          <%= f.password_field 'email.imap_password', value: AppSettings['email.imap_password'], label: "IMAP Password" %>
+        </div>
         <%= f.form_group 'email.send_email', label: { text: "Enable Outbound (SMTP) Email" }, class: 'send-email-toggle' do %>
```
scott: Could we also add a toggle and variable for SSL?  I am trying to get it working with gmail and had to make that change since they only allow SSL.

jprberlin: Sure. So toggle would set the SSL Flag. What would be the variable? And should SSL be on by default?

jprberlin: ok, i added a security setting that is either SSL or StartTLS

scott:
I must be doing something wrong, I am still having troubles getting this working with gmail.  When I run the rake task, I see the message being found, but it never shows up in Helpy.  Also there is an error line at the top:
```
bundle exec rake helpy:mailman
imap config found
#<NameError: undefined local variable or method `message' for #<Mailman::Application:0x007fa9a34731b0>>
I, [2016-05-18T07:13:33.551929 #28278]  INFO -- : Mailman v0.7.3 started
I, [2016-05-18T07:13:33.552043 #28278]  INFO -- : IMAP receiver enabled (inbound.helpy@gmail.com@imap.gmail.com).
I, [2016-05-18T07:13:33.600163 #28278]  INFO -- : Polling enabled. Checking every 60 seconds.
I, [2016-05-18T07:13:35.689596 #28278]  INFO -- : Got new message from 'andy-noreply@google.com' with subject 'Scott, welcome to your new Google Account'.
I, [2016-05-18T07:13:36.067432 #28278]  INFO -- : Got new message from 'inbox+noreply@google.com' with subject 'Welcome to Inbox by Gmail'.
I, [2016-05-18T07:17:41.596793 #28278]  INFO -- : Got new message from 'scott.miller.utah@gmail.com' with subject 'Test of IMAP integration'.
```

Any ideas?  None of the three messages shows up in my local instance of Helpy and the dev log doesn't show any kind of activity on the message which makes me think that error line may have something to do with it?

jprberlin:
I added a default scope, that fixed one issue. I also change the lookup of users, so that is not case sensitive

scott:
Nice!  I did notice some artifacts from a merge though:

screen shot 2016-05-18 at 3 39 30 pm

scott:
I saw you added a setting for the port, can you also integrated that into the settings UI/tests?

jprberlin:
What do you prefer? Dropdown with common ports or text field?

scott:
Hmm thats a good question, probably stay with a text field just in case.

scott:
Unfortunately I had to revert this- after a couple days in production on support.helpy.io I noticed a ton of exceptions coming in from the inbound mail handler (we use Sendgrid).  On further inspection I noticed the problem was here:

```
undefined method `[]' for #<Griddler::Email:0x00000009bc8690>
lib/email_processor.rb:74:in `get_email_from_mail'
```

Can you look into this and resubmit a PR for further testing?  It appears there needs to be some kind of switch here depending on whether the message is coming from `griddler` or not?

scott:
I can also provide a full rails error transcript output if that would be helpful.

jprberlin:
OK. It seems like my new code expectes an instance of Mail as its provided by mailin. Griddler seems to handle this proprietary. I will take care.

PR #149, opened 2016-05-16
Fix upvote issue

shaktikatare1989:
I have fixed the upvote issue on idea and qna page. Please have a look and let me know your thoughts.

Thanks!


scott:
Hey thanks!  This looks like it has introduced a formatting problem here.  It only happens on some of them though, but at least on some it renders the partial incorrectly, without the line "30 days ago {username} asked a question..."

screen shot 2016-05-16 at 12 23 49 pm

scott:
Hmmm looks like that formatting issue already existed, so disregard the comment above!

shaktikatare1989:
I have update code so that, only upvote counter is updated instead off whole view rendering. please review.

Issue #151, opened 2016-05-16
[Locales] key :i18n


Nondv:
[Screenshot](http://s32.postimg.org/fsa34ii8l/2016_05_16_19_22_10.png)

There're conflicts with `rails-i18n`. If you dig `I18n.config.backend` for translations you will find that there're hashes (!) in `@translations[<locale>][:i18n]`. I am not that competent to say what is that, but it generates also by [this file](https://github.com/svenfuchs/rails-i18n/blob/master/lib/rails_i18n/common_pluralizations/east_slavic.rb) (e.g. for russian locale)

Problem on screenshot is not specific for `:ru` locale, `:fr` has it too.

I am sorry for my english.

Nondv:
Created [pull request](https://github.com/helpyio/helpy/pull/152).
Just renamed i18n to international

PR #152, opened 2016-05-16
change key :i18n in locales to :international

Nondv:
There were some problems with this key, cuz some of rails-i18n files create this key.
E.g. https://github.com/svenfuchs/rails-i18n/blob/master/lib/rails_i18n/common_pluralizations/east_slavic.rb

see [this issue](https://github.com/helpyio/helpy/issues/151)


scott:
Thanks for this!  I had no idea that was causing an issue...

PR #160, opened 2016-05-18
FIX#156-The more link on the QNA view is not working

shaktikatare1989:
This was the issue with turbo-link, which protect the click event to get executed.


scott:
Thanks for tracking this down!

PR #162, opened 2016-05-18
Inactiveforumfix

jprberlin:
Do not allow creation of non private tickets if forum function is disabled


jprberlin:
sorry git issues

Issue #157, opened 2016-05-18
Add recaptcha to topics/new view if not logged in


scott:
The `topics/new` view has a different experience depending on whether the user is logged in or not.  If they are not logged in, a private discussion is created automatically, along with a user account.  If they are logged in, they can choose between a private and public discussion.

Current logged out experience:
![screen shot 2016-05-17 at 6 05 09 pm](https://cloud.githubusercontent.com/assets/14369/15343437/fade19aa-1c59-11e6-8471-27a2907516b8.png)

Current logged in experience:
![screen shot 2016-05-17 at 6 04 52 pm](https://cloud.githubusercontent.com/assets/14369/15343442/1467ea90-1c5a-11e6-890d-74ce7500f486.png)

This enhancement will change this default behavior so that an unlogged in user will have the choice between public or private discussion (same as current logged in state) and be shown a recaptcha to prevent against robot spam.  Logged in users will not see any difference in experience. Future logged out experience mockup:

![mockup](https://cloud.githubusercontent.com/assets/14369/15395024/449bbfae-1d93-11e6-96e7-ec2d3c826669.png)

To handle the secret and key, two new variables should be added to the AppSettings: `AppSettings['recaptcha.site_key']` and `AppSettings['recaptcha.api_key']` and made editable via the settings UI on the general settings panel.  

You can find this on `views/admin/settings/index.html.erb`

These new AppSettings should also be added to the `default_settings` initializer.  Once completed, it should look like this:

![recaptcha mock](https://cloud.githubusercontent.com/assets/14369/15343792/63bbad22-1c5d-11e6-8ec9-4defd6686dc0.png)

If the site owner decides not to use recaptcha, then the current version of `topics/new` should be displayed:

![screen shot 2016-05-17 at 6 05 09 pm](https://cloud.githubusercontent.com/assets/14369/15343437/fade19aa-1c59-11e6-8471-27a2907516b8.png)

shaktikatare1989:
Hi scott,

I have two questions.
1) from where site-owner decides not to use recaptcha. 
2) when any user create new topic by using recaptcha, how we build that user without its email and name?

Please let me know your thoughts.

scott:
1- That is in the admin settings of Helpy- `app/views/admin/settings/settings.html.erb`
2. Oops! It should have the email and name fields from above.  I have updated the mockup to show this.

shaktikatare1989:
[...]

PR #164, opened 2016-05-20
Worked on: Add recaptcha to topics/new view if not logged in- #157

shaktikatare1989:


scott:
Overall this works well, although I noticed one bug- the "widget" throws an error if recaptcha is enabled.  I think the widget should not include the recaptcha form due to space concerns and the fact that it never posts to the public forums.

The tests appear to to pass most of the time, but this is a bit misleading because if there were keys present for recaptcha they would fail.  Can you add tests for topic creation via widget and topic controllers when the keys are present?  Currently the tests load a method called `set_default_settings` which only creates the case where recaptcha is not present.

Hope this makes sense- but because recaptcha is optional, we need to account for the case where it is present and the case where it is not present.  I did the same thing with cloudinary configs in `topics_controller_test`

Issue #142, opened 2016-05-12
Cloudinary images are passive mixed content (non-https) 


danblah:
Seems the Cloudinary img tags are http rather than // and without a Content-Security-Policy not getting upgraded.

PR #167, opened 2016-05-24
Use secure cloudinary link

shaktikatare1989:
PR regarding https://github.com/helpyio/helpy/issues/142


scott:
The original reporter of this issue suggested using just `//res.cloudinary.com/#{Cloudinary.config.cloud_name}/image/upload` what do you think about this as opposed to hard coding `https`  I am curious what the difference is, will that not work because you have to specify secure in the config?

shaktikatare1989:
I have used secure option in config so that it will always use https instead of http when we use cl_image_tag, and on the places where there is static http used I have replace them with https to prevent passive mixed content.

PR #173, opened 2016-05-27
Refactor forms

shaktik1989:
I have re-factor the forms of the admin site, Please review.


scott:
On first glance, this looks good, but I wonder if we can figure out how to handle the checkbox validation better.  Currently it is doing something weird- it puts the checkmark over the text... wondering if there is an option (or we just need CSS) to move it over to the right of the label.

screen shot 2016-05-28 at 12 37 28 pm

shaktikatare1989:
Hi Scott,

I have added code to re-factor topic/new form, I had to create some extra commits due to travis errors.

app/views/admin/categories/edit.html.erb
```diff
@@ -7,9 +7,9 @@
 
 <%= locale_select if AppSettings['i18n.available_locales'].count > 1 %>
 
-<%= bootstrap_form_for @category, url: admin_category_path(id: @category.id), label_errors: true, method: 'patch' do |f| %>
+<%= simple_form_for @category, validate: true, url: admin_category_path(id: @category.id), label_errors: true, method: 'patch' do |f| %>
   <% Globalize.with_locale(params['lang']) do %>
-    <%= f.alert_message "Please fix the errors below." %>
+    <%= render 'shared/error_messages', object: f.object %>
```
scott: This is not a blocker but I didn't see this actually getting displayed- I think the client side validations just prevent submit if there are errors.  I guess this is in case JS is not enabled?

app/views/admin/docs/edit.html.erb
```diff
@@ -1,9 +1,9 @@
 <h2 class="form-subhead"><%= t(:edit, default:'Edit') %>: <%= @doc.title %></h2>
 <%= locale_select if AppSettings['i18n.available_locales'].count > 1 %>
-<%= bootstrap_form_for :doc, url: admin_category_doc_path(category_id: @category.id, id: @doc.id), :method => 'patch' do |f| %>
+<%= simple_form_for :doc, url: admin_category_doc_path(category_id: @category.id, id: @doc.id), :method => 'patch' do |f| %>
```
scott: I noticed this one is missing `validate: true`  Also, there is an exception on submitting invalid data:

```
undefined method `id' for nil:NilClass
Extracted source (around line #3):
         
<h2 class="form-subhead"><%= t(:edit, default:'Edit') %>: <%= @doc.title %></h2>
<%= locale_select if AppSettings['i18n.available_locales'].count > 1 %>
<%= simple_form_for :doc, url: admin_category_doc_path(category_id: @category.id, id: @doc.id), :method => 'patch' do |f| %>

  <% Globalize.with_locale(params['lang']) do %>
  <%= render 'shared/error_messages', object: f.object %>
```

app/views/admin/docs/edit.html.erb
```diff
@@ -1,9 +1,9 @@
...
 
   <% Globalize.with_locale(params['lang']) do %>
-  <%= f.alert_message "Please fix the errors below." %>
+  <%= render 'shared/error_messages', object: f.object %>
   <%= render :partial => 'form', :object => @doc, :locals => {:f => f} %>
   <% end %>
 <% end %>
```
scott: I also noticed that for some reason, all form fields are picking up an asterisk indicating they are required even though they are not, and the booleans are showing up as text_fields:

screen shot 2016-06-03 at 11 58 53 am

This seems to only occur on the edit view, not the new view.

scott:
I am going to merge anyways, but I noticed that the client side validation on the docs create/edit views is not enforcing the category_id validation.  IE, I can submit with the category_id dropdown still on "Select Category".  I will create a ticket and assign to you.  Thanks for this work!

PR #179, opened 2016-06-01
Fix mail processor so it can handle mailman and griddler

janrenz:
Hi provided a fix. However i am unable to pull master into here, as this kills a lot of my code to to your revert. maybe you can take care of this.


scott:
Yes I think I can merge it locally.

scott:
Okay cool, I got it sorted out in a staging branch and will test this in production before merging in just to make sure :)  Thanks for the patch!

scott:
It would be great if someone wanted to sort this out.  The functionality is actually working in the branch but needs to be updated and conflicts resolved.

PR #181, opened 2016-06-03
Refactor topic create

shaktik1989:
I have tried to refactor some of the code of topic#create action.


app/models/topic.rb
```diff
@@ -144,6 +144,15 @@ def public
...
 
+  def create_topic_with_user(params, current_user)
+    self.user = current_user ? current_user : User.find_by_email(params[:topic][:user][:email]) 
+    
+    unless self.user #User not found, lets build it
+      self.build_user(params[:topic].require(:user).permit(:email, :name)).signup_guest
+    end
+    self.user.persisted? and self.save
+  end  
+
```
scott: Nice refactor!  Can you add a spec or two?  I know this is probably covered by the controller tests, but it would be nice to have a model test as well.

app/models/user.rb
```diff
@@ -119,6 +118,16 @@ def temp_email(auth)
...
+  def signup_guest
+    enc = Devise.token_generator.generate(User, :reset_password_token)
+    self.reset_password_token = enc
+    self.reset_password_sent_at = Time.now.utc
+
+    self.login = self.email.split("@")[0]
+    self.password = User.create_password
+    self.save
+  end
+
```
scott: Nice. Can you write a unit test?

CGA1123:
@shaktik1989 Is this still being worked on?

There are a few merge conflicts. If not, could you close this?

scott:
This still needs to be merged in.  However, I want to also change it a little bit as mentioned in #281 so it has been waiting for a bit.

Issue #175, opened 2016-05-31
Live search


brasilikum:
Do you plan on adding Live-search to the helpdesk?
I think this would be very valuable as users could immediately see their problem if it is already known.
Also it would prevent the problem that users enter search terms that are too specific and end up not finding the right soution.

PR #183, opened 2016-06-07
Autocomplete feature

shaktik1989:
I have implemented live-search, please have a look and let me know your suggestions/improvement.
Thanks!


scott:
This looks like a great start, but it needs some view styling and UX.  Can we display results the way they are styled and rendered in the current results view?

screen shot 2016-06-07 at 4 21 39 pm

Instead of using the whole text of the result, just a link and excerpt to take them to the actual article or topic page.  I do like the idea of using arrows to navigate up and down through the results if they don't want to use the mouse.  Bind the enter key to click on the result they have highlighted.

Since we'll be clicking through to the result page, then there is no need to show any kind of text preview lower on the page either:

live-search-mockup

Probably the trickiest part of this is figuring out how to handle the responsive views. If we just make the result panel the same width as the search box, that should work in all cases:

live-search-mockup2

shaktik1989:
I have update the code as per your suggestion, please have a look.

scott:
I updated `nokogiri` in master so if you pull, travis will no longer give that security warning.

scott:
In case you missed it when I posted in slack, here is one more minor change:

screen shot 2016-06-08 at 7 46 16 am

Also, I noticed that the search preview does not search on partial words... IE if you type "inst" it will not show results for "install".  I remember this was a limitation of `pg_search` but maybe you could poke around and see if you have any ideas to make this work.

scott:
This is awesome!  I noticed one other issue with the way results are returned.  It has to do with the way multilingual searches happen.  

If you have an article that you have translated, lets say to french, and you are viewing Helpy in FR locale, then when you search, the results should appear with the translated version.  This works correctly on the full page search results format, but in the live search, the results are coming back only in english.  

I suspect the json is just not displaying the translated version.  This happens pretty much automagically via the `globalize` gem, but may need to be called explicitly for the json paylow.... just a guess.

scott:
Closes #175, Thanks @shaktik1989

PR #190, opened 2016-06-10
Fix strange behaviour in last activities inside posts

Rynaro:
I don't know if anyone notice a strange behaviour inside the list posts. Everytime that anyone inside Helpy reply a message, the activity time calculated is the same of the 'first' post.
Check my screenshots to more explanation:

Here the older code:
timeline1

The fix:
timeline2

---
Seems strange to me this behaviour, and the fix, is so simple. But I don't know if this solution is inside the Helpy code metrics. :blush:

### DIFF
--- a/app/assets/stylesheets/admin.scss
+++ b/app/assets/stylesheets/admin.scss
@@ -116,7 +116,7 @@ ul.dropdown-menu {
 }
 
 #user-tickets ul li {
-	height: 25px;
+	margin-bottom: 7px;
 	color: #666;
 }
 
--- a/app/views/admin/topics/_post.html.erb
+++ b/app/views/admin/topics/_post.html.erb
@@ -4,7 +4,7 @@
 		<%= avatar_image(post.user, size=40) %>
 	</div>
 	<div class="media-body media-right">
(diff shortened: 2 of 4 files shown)


scott:
Thanks for this @Rynaro, I need to try and remember why I did it that way in the first place, clearly it was "intentional" :)  But if I can't see any good reason I will certainly merge your in, it seems to make more sense.

Rynaro:
I've added a commit adjusting the private message list, in tickets. In CSS, we have `height: 25px`, if a ticket/post with a big title is marked as private, this ticket will overlaps others tickets/posts title below.
Example, in portuguese we have a lot of words, and our grammar is so complicated, and these factors resulting in big titles in most of cases.
I only fix this "problem" replacing height property to margin property.

scott:
@Rynaro can you post a screenshot of the change?  Interesting challenge with internationalization...

Rynaro:
For sure!
Here's a example. I've wroten titles in english, but with these images you can have an idea about this "design issue".
`height: 25px;` image:
discussions_height_issue1
`margin-bottom: 7px;` image (fix candidate):
discussions_height_issue

scott:
Oh, now I totally see what you are referring to!  Thanks for that little UX enhancement!

Issue #68, opened 2016-03-03
API? 


mario:
Hey mate, was just wondering if you have any plans for an API?

scott:
Yes!  At one point a contributor contacted me about actually building it, but he got busy with life and had to put the work on hold...

olliebennett:
I'd be interested in hearing what you planned to build on top of the API. What kind of integration are you planning?

scott:
There is another thread over on http://support.helpy.io/en/topics/51-is-there-an-api-option/posts about the same thing.

mario:
I was actually planning on building mobile apps for Helpy, and a support SDK for mobile apps on top of it ... just wanted to check if you have API before I contact you to discuss details (and my ideas) regarding that. If you've got an email that I can hit ... ^_^

olliebennett:
It would be great to have an SDK for Android/iOS which allowed app users to have conversations with the support team through Helpy.

I also think there's a lot of value in a (perhaps hybrid) app for the Admin interface, which would provide the support team with updates to their assigned topics etc, allowing for more timely responses.

scott:
@olliebennett I agree on both- in fact the admin app/SDK is something that I had in the back of my mind from the beginning.  @mario We have a slack room set up, shoot me an email at hello at scottmiller.io and we can discuss this more offline.

jmarnold:
Loving all of the work y'all have done so far! Just curious, how in-depth is the API work you have planned? The one killer feature I'm looking around for is Slack integration since my organization thrives on it. Is that anywhere in your plans yet?

scott:
@jmarnold Slack integration is definitely something I have been thinking of.  Slack integration can mean a few things- are you thinking more of getting alerts in Slack when there is a new discussion, being able to reply to discussions through slack, or both?

jmarnold:
@scott Oh man, I wasn't thinking of going too crazy ;). I was just thinking of some basic notifications. My organization uses Slack as kind of the hub of ALL information so being able to push stuff there would be HUGE.

mbajur:
[...]

PR #193, opened 2016-06-14
[QA] Add JSON API with Grape

scott:
Reference #68


scott:
@wildtangent @mario @olliebennett  wondering your thoughts on if you think the API should be limited to agents/admin, or should it give all users access via oauth2?  There would be some basic permissions to limit `users` to things like viewing `docs` and their own `topics`, but i am having trouble deciding on the value of this.

olliebennett:
I'm thinking about the type of use-cases and what sort of access they'd require. My thoughts, in order of how useful / important the use-case is:

- Agent Access: Access for agents and admins to reply to (and close etc.) topics through other interfaces, such as via a "Helpy Admin" mobile app pointed at their Helpy instance.
- Automatic Ticket Creation: An authenticated service with ticket-creating and read-all permissions which is not representing an actual "user" but - for instance - another system or server. For example, an auto-importer from an existing CRM tool, a statistics extraction tool for reporting key company metrics into a central dashboard (like Geckoboard), or perhaps a two-way chat (Slack) interface allowing all employees to interface with Helpy without having their own accounts(?).
- User Registration: Creating a topic directly from an app or website (e.g. @mario's SDK idea); the user can register themselves (hopefully done automatically if we know their details already) and then open their own tickets.
- Unauthenticated Access: An unauthenticated user (site visitor or app user) should be able to browse (and search) all public topics, perhaps as a way to consume a "FAQ" section of the app.

Based on these use-cases, I think it's a good idea to begin with limiting to admins only if that keeps things simpler, and while the API is tested. As use-cases are developed, additional functionality can be exposed as required.

connorshea:
I should note that for full support of Rails 5 you'll want 0.17.0. Also that drops explicit support for Ruby < 2.2.2, though it _should_ still work the test suite isn't run for it.

PR #198, opened 2016-06-18
add simple form and clientsidevalidation to forgot password

shaktik1989:


scott:
I was actually thinking of this page: http://helpy.local/en/users/password/edit?reset_password_token=any1pzbPYUfaxY1jmiTU

It does not have any client side validations

screen shot 2016-06-18 at 1 00 24 pm

scott:
@shaktik1989 did you see this?

Issue #197, opened 2016-06-16
Add checkbox to reply admin interface to "reply and close" topic


scott:
- The checkbox should be unchecked by default
- When an admin checks the box, it will post the response to the thread, and then change the topic to "closed".  
- The system should also post an internal note to the thread indicating that the topic has been closed. (This is standard when closing topics.)
- Please include a controller spec for the behavior
- Use an i18n token for the text "Mark as resolved"

![screen shot 2016-06-15 at 8 16 25 pm](https://cloud.githubusercontent.com/assets/14369/16103615/2d2cff4a-3337-11e6-9597-2b149cd45670.png)

PR #199, opened 2016-06-18
Reply and close feature

shaktik1989:
Implementation of https://github.com/helpyio/helpy/issues/197, please have a look and let me know your thoughts.

Thanks!

### DIFF
--- a/app/controllers/admin/posts_controller.rb
+++ b/app/controllers/admin/posts_controller.rb
@@ -30,8 +30,8 @@ class Admin::PostsController < Admin::BaseController
     respond_to do |format|
       if @post.save
         format.js {
-          fetch_counts
           @topic.close(current_user.id) if params[:post][:resolved] == "1"
+          fetch_counts
           @posts = @topic.posts.chronologic
           @admins = User.agents
           #@post = Post.new
--- a/app/controllers/admin/posts_controller.rb
+++ b/app/controllers/admin/posts_controller.rb
@@ -31,7 +31,7 @@ class Admin::PostsController < Admin::BaseController
       if @post.save
         format.js {
           fetch_counts
-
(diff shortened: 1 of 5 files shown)


olliebennett:
The code looks good. The only improvement could potentially be a separate button for "Reply & Close" alongside "Reply" (as that makes this feature one-click). That would (of course) involve some JavaScript to set your "resolved" flag and the controller would not change. Either way, this feature looks good-to-go!

arrowcircle:
@olliebennett JS is not necessary. You can put custom submit parameter on the button and put resolved flag in the controller

olliebennett:
Sorry, deleted my previous comment:

> The code looks good. The only improvement could potentially be a separate button for "Reply & Close" alongside "Reply" (as that makes this feature one-click). That would (of course) involve some JavaScript to set your "resolved" flag and the controller would not change. This could be saved for a future improvement. Either way, this feature looks good-to-go!

If it's easy to do then that might be a smoother UX, but this PR is an improvement as-is regardless.

scott:
My thoughts:  @olliebennett I went back and forth on whether a button or a checkbox was a better experience just because it is too easy to click the wrong button. Yes the checkbox adds a step, but it forces a conscious decision. I think it really depends how often you will both reply and close at the same time.  A future enhancement I have in mind is an "autoclose after N number of days" so old messages get handled automatically which further reduces the need to "reply and close".

scott:
@shaktik1989 Looks good, one thing I noticed is that the stat boxes at the top don't update when the message is marked resolved (ie. the resolved count should +1 and the corresponding counter should decrease by one).  You may just be missing the `get_counts` invocation which refreshes the stats or just didn't refresh that part of the page?

PR #202, opened 2016-06-22
Add .ruby-gemset .ruby-verions and .rvmrc to gitignore

arrowcircle:
Hey!
Usually, these files to select ruby version and gemset are added to gitignore.
I added them to gitignore.

Next question is should we fix ruby version in Gemfile? 2.2.1 seems outdated. Should we remove it?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,7 @@ config/settings_bu.yml
 
 *.eml
 .DS_Store
+
+/.ruby-version
+/.ruby-gemset
+/.rvmrc


arrowcircle:
@scott Any ideas?

scott:
Hi sorry for the delay, have been a little slammed.  I think we should upgrade.  The original reasoning behind staying at 2.2.1 was that is what the Digital Ocean rails image was at.  Now I am not sure if thats still the truth and I don't care anymore either.

PR #203, opened 2016-06-28
Fix ApiKey test suite and make the whole test suite pass again

wildtangent:
Resolved a few issues with the test suite which were causing it to fail in CI

* Updated fixture definitions to use the current DB schema
* Fixed issue with `before_create` hook not generating the `access_token` before validation occurred (switched to `before_validation :generate_access_token on: :create`
* Added 100% test coverage for ApiKey model
* Fix a couple of Rubocop linting errors (line indenting, unused assigned variable)

Perhaps should add validation to ensure ApiKey is always assigned to a user?

### DIFF
--- a/app/controllers/api/v1/topics.rb
+++ b/app/controllers/api/v1/topics.rb
@@ -73,7 +73,7 @@ module API
             user_id: params[:user_id],
             private: true
           )
-          post = topic.posts.create!(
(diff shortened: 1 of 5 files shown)


scott:
Thanks for writing these tests and making the other changes.  I agree about making sure the ApiKey is always assigned to a user.

PR #204, opened 2016-06-28
Remove puts in favour of Ruby.logger.info

wildtangent:
Test runs get some `puts` statements thrown in from the `email_processor.rb` class.

This commit simply changes `puts` into `Rails.logger.info` instead to keep the test output clean while leaving the info visible during development.


scott:
Thanks for sending this!

PR #212, opened 2016-06-30
[API] Fix docs sorting and add limit option

wildtangent:
Ensure categories and docs are returned sorted by rank.
Eager load translations to save on DB queries
Add `docs_limit` option to limit the number of docs returned in the nested entities


scott:
Nice work although it now has merge conflcts

PR #208, opened 2016-06-30
[API] Re-enable Doorkeeper::Grape::Helpers...

wildtangent:
...so authentication works correctly with the authorization code in the base class

I'm guessing you wanted to add some more work around the doorkeeper/OAuth functionality. For my purposes the ApiKey token is just fine...

This gets the API working again without too much hassle. You can pass in a param of `token=` or `X-Token` as a header

EDIT: 
* Added test coverage for Categories API controller
* Added test coverage for Docs API controller
* Changed `Simplecov.start` to `Simplecov.start 'rails'`

PR #213, opened 2016-06-30
[API] Add category option to docs entity

wildtangent:
Allow category object to be returned as part of the doc object, if `category=true` is passed in as a param


scott:
Great PR, although this one has conflicts after I merged #208

PR #214, opened 2016-06-30
[API] Add docs search functionality

wildtangent:
Add functionality to support searching using the existing PG multisearch feature, but limited to just Docs.
This was a requirement in my case, might not be for everyone...

Pass in `/api/v1/docs/search?q=mysearch`

### DIFF
--- a/app/controllers/api/v1/base.rb
+++ b/app/controllers/api/v1/base.rb
@@ -12,6 +12,7 @@ module API
       mount API::V1::Topics
       mount API::V1::Posts
       mount API::V1::Users
+      mount API::V1::Search
       add_swagger_documentation
     end
   end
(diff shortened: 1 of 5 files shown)


scott:
I would prefer we (and had planned to) expose this through a search endpoint instead- and there you could offer something like this:

`/api/v1/search?type=docs&q=mysearch`
`/api/v1/search?type=topics&q=mysearch`
`/api/v1/search?q=thiswouldsearchboth`

Thoughts?

wildtangent:
I've updated this PR

Tests are a little thin at the moment but I am out of time.

I've made it so if the results are mixed, then we return a `Entity::Search` entity otherwise return the correct entity for the `type=` attribute, e.g. `Entity::Doc`, `Entity::Topic` etc

Accepts `type=Doc` and `type=docs`

scott:
Strong work! Thanks for refactoring it to be a little more flexible :)

PR #217, opened 2016-07-01
french translation

martin-letellier:
improve the french translation


scott:
Thanks so much for doing this!

PR #219, opened 2016-07-01
Spanish locale improvements

jcalonso:
Some small fixes to the Spanish locale, added missing keys


scott:
Awesome! Thanks for these :)

PR #220, opened 2016-07-03
Bulk invite agents

shaktik1989:


Gemfile
```diff
@@ -127,6 +129,7 @@ group :development, :test do
   gem 'brakeman', require: false
   gem 'rubocop'
   gem 'scss-lint'
+  gem 'awesome_print'
```
olliebennett: Is this required for this PR or just something you like to use to debug?

app/views/admin/settings/index.html.erb
```diff
@@ -16,6 +16,7 @@
         <%= settings_item('glyphicon glyphicon-modal-window', "widget", "Find out how to deploy the Helpy widget on your site.") %>
         <%= settings_item('glyphicon glyphicon-envelope', "email", "Set up SMTP and Inbound email for your Helpy.") %>
         <%= settings_item('glyphicon glyphicon-cloud', "cloudinary", "Set up Cloudinary to enable your Helpy to accept image attachments.") %>
+        <%= settings_item('glyphicon glyphicon-user', "invite", "invite agent in bulk via email.") %>
```
olliebennett: Grammar: 

> invite agent in bulk via email.

becomes

> Invite agents in bulk via email.

app/controllers/application_controller.rb
```diff
@@ -137,4 +139,10 @@ def theme_chosen
     end
   end
 
+  protected
+
+  def configure_permitted_parameters
```
olliebennett: Does this have to live in `application_controller` or can it move to a child controller elsewhere?

app/controllers/application_controller.rb
```diff
@@ -9,6 +9,8 @@ class ApplicationController < ActionController::Base
   before_action :set_vars
   before_action :instantiate_tracker
 
+  before_filter :configure_permitted_parameters, if: :devise_controller?
```
olliebennett: As RuboCop has complained, this needs to be `before_action`.

app/models/user.rb
```diff
@@ -141,4 +150,22 @@ def is_editor?
     %w( editor agent admin ).include?(self.role)
   end
 
+  def self.bulk_invite(emails, message)
+    #below line merge comma saperated emails as well as emails saperated by new lines
+    emails = emails.each_line.reject { |l| l =~ /^\s+$/ }.map { |l| l.strip.split(', ') }.flatten
+    
+    emails.each do |email|
+      if email.match('^.+@.+$')
```
olliebennett: As RuboCop has complained:

```
Use =~ in places where the MatchData returned by #match will not be used.
```

olliebennett:
This is a neat feature and will help new users to onboard quickly - nice one! :)

I've not checked whether it works yet.

Gemfile
shaktik1989: I have added this in the development environment, forgot to remove it. Will remove this in next commits.

app/controllers/application_controller.rb
shaktik1989: hmm, as per the devise docs to permit additional attribute we need to permit then like this in application controller.

app/models/user.rb
```diff
@@ -141,4 +150,23 @@ def is_editor?
...
+        end
+      end
+    end
+  end
+  
+  #when using deliver_later attr_accessor :message becomes nil on mailer view
+  def send_devise_notification(notification, *args)
+    devise_mailer.send(notification, self, *args).deliver
+  end
+
```
scott: I saw one response about this on SO suggesting to serialize the message into the job, which seems right to me. Any thoughts @shaktik1989?

config/locales/devise_invitable.en.yml
```diff
@@ -0,0 +1,31 @@
...
+        someone_invited_you: "Someone has invited you to %{url}, you can accept it through the link below."
+        accept: "Accept invitation"
+        accept_until: "This invitation will be due in %{due_date}."
+        ignore: "If you don't want to accept the invitation, please ignore this email.<br />Your account won't be created until you access the link above and set your password."
+  time:
+    formats:
+      devise:
+        mailer:
+          invitation_instructions:
+            accept_until_format: "%B %d, %Y %I:%M %p"
```
scott: Can you grab the other locale files and add them into the config as well?

app/views/devise/invitations/new.html.erb
```diff
@@ -0,0 +1,11 @@
+<h2><%= t "devise.invitations.new.header" %></h2>
+
+<%= simple_form_for resource, as: resource_name, url: invitation_path(resource_name), html: { method: :post} do |f| %>
+  <%= devise_error_messages! %>
+
+<% resource.class.invite_key_fields.each do |field| -%>
+  <%= f.input field %>
+<% end -%>
+
```
scott: I noticed this still renders although I did not test if it actually works.  I would say we have two options here- either disable it completely or lock it down to admins only.  That being said, your form that takes multiple email addresses is definitely nicer.

Gemfile
```diff
@@ -99,6 +99,8 @@ gem 'griddler-sparkpost'
 
 gem 'rails-timeago'
 
+gem 'devise_invitable', '~> 1.5.2'
+
```
scott: @shaktik1989 I discovered this needs `~> 1.6` to work with Rails 4.2.7 and Rails 5

Issue #225, opened 2016-07-12
Move Google Analytics calls to background worker?


scott:
Need to check how feasible this is, but the backend of Helpy uses  many server side MP (measurement protocol) calls to GA.  Moving these to workers should be easy and could yield a nice speed up.

Issue #222, opened 2016-07-08
Move bulk operations to background worker


scott:
Currently if you select a bunch of messages and perform a bulk action, it is really slow.  This should be offloaded to a worker.

scott:
Hi @shaktik1989 I didn't have time to write up any additional requirements of the API work yet, but you can take a look at this issue instead.  If you look in https://github.com/helpyio/helpy/blob/master/app/controllers/admin/topics_controller.rb

There are actions for changing assignee, privacy, status, etc.  This provides the ability to update the status of one (or a group of) discussions.  They all accept an array of params and iterate through topics one by one.  This can get quite slow for bulk operations- when an action is taken on many topics at once.  It would be great to offload this to a worker.  It also mean the JS should still update the UI independent of the model data having actually changed since they will now be decoupled.

olliebennett:
Would it be more appropriate to improve the performance of the queries involved instead? Seems that moving to a background worker is just working around the fact we're doing inefficient things, and also leads to potential discrepancies between front and back-end if the UI updates without confirmation that things have changed server-side (i.e. if the job fails).

To take your `admin/topics_controller.rb` example, there's some very low-hanging fruit. Below are some examples from the `assign_agent` method:

- Multiple `.find` queries instead of one bulk select
    ```ruby
    # before
    params[:topic_ids].each do |id|; @topic = Topic.where(id: id).first
    # after
    Topic.where(id: params[:topic_ids]).each do |topic|
    # note: this no longer crashes if an invalid topic ID is provided, which may or may not be desired.
    ```

- We're running `User.find(params[:assigned_user_id])` inside a loop but it's not changing (so we should calculate it before the loop).

- If `unless params[:assigned_user_id].blank?` is true, we probably don't need to bother looping through anything?

- `@topic.assign` is re-calculating the message (for i18n) again re-calculating `User.find(assigned_to).name` each time; we could potentially optimise this as an `update_all` (or other bulk create / update logic) if we were clever about it, or at least avoid instantiating the `@topic` objects. That may be going further than necessary though.

- Should we use `@count` instead of `params[:topic_ids].count > 1` or is that deliberate? If deliberate, we could check `Topic.where(id: params[:topic_ids])` against `params[:topic_ids].length` instead?

- `@count = @count + 1` => `@count += 1`

- Also it seems there are a few unnecessary instance variables (`@minutes` is only ever 0).

[...]

PR #228, opened 2016-07-14
Add worker for sending trackers to GA

scott:
This moves the server side GA measurement protocol tags to a worker. @shaktik1989 @olliebennett 

Reference: #222 #225

### DIFF
--- a/app/controllers/admin/posts_controller.rb
+++ b/app/controllers/admin/posts_controller.rb
@@ -37,9 +37,11 @@ class Admin::PostsController < Admin::BaseController
           #@post = Post.new
           case @post.kind
           when "reply"
-            @tracker.event(category: "Agent: #{current_user.name}", action: "Agent Replied", label: @topic.to_param) #TODO: Need minutes
+            # @tracker.event(category: "Agent: #{current_user.name}", action: "Agent Replied", label: @topic.to_param) #TODO: Need minutes
+            tracker("Agent: #{current_user.name}", "Agent Replied", @topic.to_param, nil) #TODO: Need minutes
           when "note"
-            @tracker.event(category: "Agent: #{current_user.name}", action: "Agent Posted Note", label: @topic.to_param) #TODO: Need minutes
+            # @tracker.event(category: "Agent: #{current_user.name}", action: "Agent Posted Note", label: @topic.to_param) #TODO: Need minutes
+            tracker("Agent: #{current_user.name}", "Agent Posted Note", @topic.to_param, nil #TODO: Need minutes
           end
           render 'admin/topics/show'
         }
--- a/app/controllers/admin/posts_controller.rb
+++ b/app/controllers/admin/posts_controller.rb
@@ -30,7 +30,10 @@ class Admin::PostsController < Admin::BaseController
     respond_to do |format|
       if @post.save
         format.js {
-          @topic.close(current_user.id) if params[:post][:resolved] == "1"
+          if params[:post][:resolved] == "1"
+            @topic.close(current_user.id)
+            tracker("Agent: #{current_user.name}", "Closed", @topic.to_param) #TODO: Need minutes
+          end
           fetch_counts
           @posts = @topic.posts.chronologic
           @admins = User.agents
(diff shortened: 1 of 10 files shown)


app/jobs/tracker_job.rb
```diff
@@ -0,0 +1,32 @@
+class TrackerJob < ActiveJob::Base
+  queue_as :default
+
+  def perform(category, action, label, value, client_id, ga_id)
+    if ga_id.present?
```
olliebennett: This check perhaps isn't necessary given you're checking `if AppSettings['settings.google_analytics_id'].present?` in your `tracker` method? No harm double checking though, I suppose.

olliebennett:
Nice one - this looks good. It would be cool to extend it so that the job could accept an array of events to report; that would be especially handy for bulk processing and would reduce the overhead of adding and processing many jobs; probably not worth the complexity though. This alone will greatly speed things up so it gets a 👍  from me!

app/controllers/admin/posts_controller.rb
```diff
@@ -37,9 +37,11 @@ def create
           #@post = Post.new
           case @post.kind
           when "reply"
-            @tracker.event(category: "Agent: #{current_user.name}", action: "Agent Replied", label: @topic.to_param) #TODO: Need minutes
+            # @tracker.event(category: "Agent: #{current_user.name}", action: "Agent Replied", label: @topic.to_param) #TODO: Need minutes
```
olliebennett: Can we delete the comments?

app/controllers/admin/search_controller.rb
```diff
@@ -20,22 +20,23 @@ def topic_search
     if users.size == 0 # not a user search, so look for topics
       @topics = Topic.admin_search(params[:q]).page params[:page]
       template = 'admin/topics/index'
-      @tracker.event(category: "Admin Search", action: "Topic Search", label: params[:q])
+      tracker("Admin Search", "Topic Search", params[:q], nil)
```
olliebennett: Is the `nil` necessary? It's implied if we don't supply it, AFAIK.

app/controllers/application_controller.rb
```diff
@@ -39,6 +39,19 @@ def verify_admin
     (current_user.nil?) ? redirect_to(root_path) : (redirect_to(root_path) unless current_user.is_admin?)
   end
 
+  def tracker(ga_category, ga_action, ga_label, ga_value=nil)
```
olliebennett: likewise here, the `=nil` is redundant as that's default if the param is not supplied anyway.

app/controllers/posts_controller.rb
```diff
@@ -51,7 +51,8 @@ def create
           @posts = @topic.posts.ispublic.chronologic.active
           unless @topic.assigned_user_id.nil?
             agent = User.find(@topic.assigned_user_id)
-            @tracker.event(category: "Agent: #{agent.name}", action: "User Replied", label: @topic.to_param) #TODO: Need minutes
+            # @tracker.event(category: "Agent: #{agent.name}", action: "User Replied", label: @topic.to_param) #TODO: Need minutes
```
olliebennett: Again - should we scrap these comments?

app/jobs/tracker_job.rb
scott: Partly a result of a refactoring, but I left it in just because there are potential use cases for calling the job directly.. such as using a different ga_id.

app/controllers/admin/search_controller.rb
scott: Again, this was a refactor that didn't get fully cleaned up.

app/controllers/application_controller.rb
scott: I am not sure about that, in my testing it seemed to not log the event without setting to nil explicitly.

olliebennett: Sorry for the confusion - you're right that not passing the correct number of arguments leads to `ArgumentError: wrong number of arguments` unless you set the default. I guess what I mean was there's no need to combine passing `nil` as a parameter when calling `tracker()` _as well as_ defaulting the param to `nil` as you're doing here; it seems superfluous. No harm done though, and the PR gets my 👍  with or without this! :)

scott:
Closes #225

PR #229, opened 2016-07-14
[WIP] Configure OpenShift

olliebennett:
The QuickStart I created:
https://hub.openshift.com/quickstarts/298-helpy

Deploy link for the QuickStart (which the button would link to):
https://hub.openshift.com/quickstarts/deploy/298-helpy

I'm currently having problems with the deployment (seems hard to debug without the full logs) but will try again later. Not sure what will be involved to get it fully working, but it seems to be getting quite far.

### DIFF
--- a/.openshift/README.md
+++ b/.openshift/README.md
@@ -1,0 +1,3 @@
+For information about .openshift directory, consult the documentation:
+
+http://openshift.github.io/documentation/oo_user_guide.html#the-openshift-directory
--- a/.openshift/action_hooks/README.md
+++ b/.openshift/action_hooks/README.md
@@ -1,0 +1,3 @@
+For information about action hooks, consult the documentation:
+
+http://openshift.github.io/documentation/oo_user_guide.html#action-hooks
--- a/.openshift/action_hooks/build
+++ b/.openshift/action_hooks/build
@@ -1,0 +1,5 @@
+#!/bin/bash
+# This is a simple build script and will be executed on your CI system if
+# available.  Otherwise it will execute while your application is stopped
+# before the deploy step.  This script gets executed directly, so it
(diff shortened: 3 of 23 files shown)


olliebennett:
Next issue:

```
Your Ruby version is 2.0.0, but your Gemfile specified 2.2.1
```

... for now I've removed the ruby version lock from `Gemfile`, because later Ruby versions are not yet supported. [This page](https://www.openshift.com/features/cartridges.html) suggests that `Ruby 2.0.0` is "Online (Next Gen)" which is promising, although there seems to be no cartridge available named `ruby-2.2`...

olliebennett:
We need to make a decision about which version of the OpenShift platform we want to use.

The legacy version (v2?) seems to be the default; examples are for [Rails 4](https://github.com/openshift/rails4-example) (or [Rails 3](https://github.com/openshift/rails-example)), and it's almost working, although the Ruby version is locked to 2.0 (and unlikely to be updated to 2.2 on the legacy platform I would assume).

The new (containerised) version (v3?) has a Rails 4 example [here](https://github.com/openshift/rails-ex), but seems more complicated to manage, and certainly seems more complicated to set up initially, and I don't see any reference to the "OpenShift Button" idea.

@brianjking do you have any other ideas?

olliebennett:
If OpenShift doesn't meet our needs, we could alternatively create an [AWS Marketplace](https://aws.amazon.com/marketplace) offering, which should be pretty quick for users to get up and running.

scott:
Thanks for digging into this @olliebennett! It seems like their implementation is not quite as easy to work with as Heroku.  We could also create and maintain an "official" openshift branch and try to backport to Ruby 2 but I don't see a lot of value there.

PR #231, opened 2016-07-16
Replace 'quiet_assets' gem with config

olliebennett:
Depends on #230.

This PR simply removes deprecated [quiet_assets gem](https://github.com/evrone/quiet_assets) replacing it with a config option as documented in their [readme](https://github.com/evrone/quiet_assets#deprecation).


olliebennett:
Rebased! :)

PR #232, opened 2016-07-16
Remove RSpec

olliebennett:
Although this gem was present, it was not being used, and there were no "specs" in the codebase.

Out Minitest test suite is quite extensive and fit for purpose, and although switching to RSpec could be achieved, I don't see any compelling reason to do so urgently.

Could always re-introduce this gem later if needed.


scott:
Yeah It think the rspec dates back to when I was still deciding which direction to go.  Safe to remove :)

PR #238, opened 2016-07-16
refactor assign agent method

shaktik1989:
@olliebennett , @scott : I have tried to re-factor  agent invite method, please review and give suggestions and feedback. Thanks!


app/controllers/admin/topics_controller.rb
```diff
@@ -182,36 +182,40 @@ def update_topic
...
-      @count = @count + 1
-    end
+    end 
+    
+    @topics.bulk_assign(bulk_post_attributes, assigned_user.id) if bulk_post_attributes.present?
 
     if params[:topic_ids].count > 1
       get_tickets
     else
+      @topic = @topics.first
```
olliebennett: Subtle change in behaviour here: previous code would be equivalent to `@topic = @topics.last` I think? Or - at least - it would have been the topic for the last ID in `params[:topic_ids]` (where the sort order is unspecified?). Not sure whether this is important anyway as don't know what `@topic` is used for.

app/controllers/admin/topics_controller.rb
```diff
@@ -182,36 +182,40 @@ def update_topic
...
+        bulk_post_attributes << {body: I18n.t(:assigned_message, assigned_to: assigned_user.name), kind: 'note', user_id: previous_assigned_id, topic_id: topic.id}
+        #topic.assign(previous_assigned_id, assigned_user.id)
 
         # Create internal note
-#        @topic.posts.create(user_id: previous_assigned_id, body: "Discussion has been transferred to #{assigned_user.name}.", kind: "note")
+        # @topic.posts.create(user_id: previous_assigned_id, body: "Discussion has been transferred to #{assigned_user.name}.", kind: "note")
 
         # Calls to GA
-        @tracker.event(category: "Agent: #{current_user.name}", action: "Assigned to #{assigned_user.name.titleize}", label: @topic.to_param, value: @minutes)
+        @tracker.event(category: "Agent: #{current_user.name}", action: "Assigned to #{assigned_user.name.titleize}", label: topic.to_param, value: @minutes)
```
olliebennett: This will conflict with Scott's new PR (#225) once that's merged, so will need to be dealt with later.

Is the `@minutes` variable used for something special, or can we just pass `0` in here directly? I don't see where it's changed otherwise...

app/controllers/admin/topics_controller.rb
```diff
@@ -182,36 +182,40 @@ def update_topic
...
-        previous_assigned_id = @topic.assigned_user_id? ? @topic.assigned_user_id : params[:assigned_user_id]
-        assigned_user = User.find(params[:assigned_user_id])
-        @topic.assign(previous_assigned_id, assigned_user.id)
+        previous_assigned_id = topic.assigned_user_id || params[:assigned_user_id]
+        bulk_post_attributes << {body: I18n.t(:assigned_message, assigned_to: assigned_user.name), kind: 'note', user_id: previous_assigned_id, topic_id: topic.id}
+        #topic.assign(previous_assigned_id, assigned_user.id)
 
         # Create internal note
-#        @topic.posts.create(user_id: previous_assigned_id, body: "Discussion has been transferred to #{assigned_user.name}.", kind: "note")
+        # @topic.posts.create(user_id: previous_assigned_id, body: "Discussion has been transferred to #{assigned_user.name}.", kind: "note")
```
olliebennett: Great to have removed this; might as well delete completely instead to clean up old comments?

app/models/topic.rb
```diff
@@ -126,6 +126,11 @@ def assign(user_id=2, assigned_to)
     self.save
   end
 
+  def self.bulk_assign(post_attributes, assigned_to)
+    Post.create(post_attributes)
```
olliebennett: This is great; nice one for using `create` with an array - that's a good tidy up even though I think `create` just loops through the array internally (so still does one query for each insert).

app/models/topic.rb
```diff
@@ -126,6 +126,11 @@ def assign(user_id=2, assigned_to)
...
+    self.update_all(assigned_user_id: assigned_to, current_status: 'pending')
```
olliebennett: This seems like a (big) problem; are you sure that `self.update_all` will not update _all_ topics (not just those which were created)?

I think a it's necessary to do:

```ruby
new_posts = Post.create(post_attributes)
new_posts.update_all(assigned_user_id: assigned_to, current_status: 'pending')
```

... am I correct?

olliebennett: I don't know what happens if any one of the posts fails to create (for any reason) - it would be good to think about what happens / what should happen in this case, perhaps?

olliebennett: I initially panicked that `update_all` would affect _all_ topics, but see you've called this on `@topics` specifically, so it will only apply to those.

app/models/topic.rb
```diff
@@ -126,6 +126,11 @@ def assign(user_id=2, assigned_to)
     self.save
   end
 
+  def self.bulk_assign(post_attributes, assigned_to)
```
olliebennett: Can we remove the old `assign` method now we're using this instead? It seems it's not used elsewhere (other than it tests).

olliebennett:
I really like this! Am looking forward to seeing what kind of performance impact it has.

We could do with a test for `bulk_assign` I suppose, but otherwise I say good job 👍

app/controllers/admin/topics_controller.rb
shaktik1989: I have thought that, when it comes inside else in that case we have only one topic in @topics. so its always the save topic when we use first or last.

app/controllers/admin/topics_controller.rb
shaktik1989: yes we can pass 0 directly, will update this.

app/models/topic.rb
shaktik1989: yes, do you have any thoughts how to make it in single query?

app/controllers/admin/topics_controller.rb
scott: I just merged in the GA tracker work I have done, so you will see that if you pull from master.

app/models/topic.rb
scott: I suspect if we could figure out how to do a true bulk insert (instead of looping through) that would shave quite a bit of time.  If not, just this action alone could be a candidate for background processing if the number of `topics` is greater than 1 (ie. we are on the list view).

scott:
I did some "poor mans" profiling just by looking at the times in the dev mode server output and didn't see a great improvement in time. Does anyone know if this is an accurate way to profile speeds?  I just selected 25 messages on the list view and reassigned them and then looked at the time.  I am guessing the message insert (mentioned above) is taking the bulk of the time, since it loops through and does 25 inserts...

scott:
WOW! What a difference that made.  It absolutely flies now!  I did notice that the JS update problem still exists, here is a screen cap to illustrate what I am talking about.  The dropdown label should update with the assigned agents name:

single_assign

olliebennett:
Haha - awesome! This really is great; I love these kind of performance improvements, and the `bulk_insert` gem seems much cleaner than other ways I've seen... well done @shaktik1989 ! 👍

app/controllers/admin/topics_controller.rb
```diff
@@ -202,7 +202,7 @@ def assign_agent
     if params[:topic_ids].count > 1
       get_tickets
     else
-      @topic = @topics.first
+      @topic = Topic.find(@topics.first.id)
```
olliebennett: That's an interesting bug. Although I didn't consider it before, it makes sense that the `@topics.first` has been modified by a direct query so ActiveRecord is not able to update the in-memory version.

A neater way to achieve this (although it does the same thing under the hood) is to use

```
@topic = @topics.first.reload
```

olliebennett:
I think this is good to go, isn't it? I've added a note about using `@topic = @topics.first.reload` but it's a minor code style tweak, so I vote :shipit: whether you change it or not :)

Once again @shaktik1989 this is awesome work! ;)

scott:
I agree, I am going to give it one more scrub, but I think its ready and is a great contribution by @shaktik1989

PR #240, opened 2016-07-16
Use AppSettings value for mailers host value

scott:


olliebennett:
Cool - does this work correctly with protocols defined? For example, in the `production.rb` in our codebase, we have:

```ruby
  # ActionMailer Config
  # Setup for production - deliveries, no errors raised
  config.action_mailer.default_url_options = {
    host: 'www.example.com',
    protocol: 'https'
  }
```

... is there any way the user can break things if they add the protocol (eg. `http`) to their AppSettings' `settings.site_url`?

scott:
Thats a good question, probably worth testing out.

PR #246, opened 2016-07-21
Remove unreachable code

joshuaclayton:
Hello!

I've been building a tool to [identify unused code](https://unused.codes/) and found Helpy on http://www.opensourcerails.com/.

Each discrete commit outlines removal of unused code, as well as the corresponding commit(s) that led up to it becoming unreachable. I've verified the test suite runs green for each commit; however, I'm obviously not as familiar with the codebase so if there are areas where metaprogramming are involved or integration tests don't cover certain aspects, please let me know and I can adjust.

Thanks!

### DIFF
--- a/app/models/topic.rb
+++ b/app/models/topic.rb
@@ -86,10 +86,6 @@ class Topic < ActiveRecord::Base
     "##{self.id} | #{self.name}"
   end
 
-  def open?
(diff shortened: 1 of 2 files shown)


scott:
Thats a really interesting tool, great for removing cruft.  I will take a look at these suggestions.

Issue #224, opened 2016-07-12
Rails 5


connorshea:
I wanted to suggest upgrading Helpy to Rails 5, or at least get a conversation started about it.

I've been keeping track of our progress getting to the upgrade at [GitLab](https://gitlab.com/gitlab-org/gitlab-ce/issues/14286), figured you may find all my notes useful :)

Some notes based on a quick look through your Gemfile:

- Devise needs to be upgraded to 4.0+
- quiet_assets can be replaced by a sprockets-rails config option now (wait until Rails 4.2.7 for that, though)
- Rspec will need to be updated to 3.5.0
- acts-as-taggable-on is a blocker at the moment
- kaminari should be updated to 0.17.0
- route_translator needs to be upgraded to 5.0.0
- sass-rails needs 5.0.5
- globalize needs a new version released

Lots of information on upgrading those gems, etc. is available in the issue I linked above. This project looks awesome, kudos!

Best,
Connor

scott:
Thanks Connor,  your tracking of this is really appreciated.  I have been tracking Globalize since thats a really big blocker for us to go to Rails 5.  I plan to start an official branch where we can start to work out the upgrade.

connorshea:
:+1: to my knowledge we're the largest open source Rails app and have somehow managed to not fall behind on Rails releases, so I figured it'd be useful to track the progress in-depth for everyone else to learn from :)

Feel free to ping me if you run into any problems, I might be able to help - or at least know to look out for the problem in GitLab!

olliebennett:
Great - thanks for kicking this off, @connorshea.

@scott it may be worth setting up a GitHub milestone for this. Until then, I'll track the related PRs here:

#230 - Update 'rails' and dependencies to 4.2.7

scott:
As you can see, I set up a milestone.  I will start to tick off these PRs as time allows.  Most hopefully most will go in fairly easily.

connorshea:
@scott @olliebennett with Devise you can just update to 4.0 for now, which has Rails 5 support and just a deprecation warning for the omniauth change.

See https://github.com/plataformatec/devise/pull/3985

connorshea:
You'll also need to upgrade the Gemfile to use at least Ruby 2.2.2 (right now it's 2.2.1), I'd recommend the latest patch release in the 2.2 series.

scott:
Thanks @connorshea

connorshea:
[...]

PR #252, opened 2016-07-21
Update Ruby version to latest (2.2.5)

olliebennett:
Rails 5 depends on Ruby >= 2.2.2. I'm updating here to the latest available in the 2.2.x series as suggested in #224.

We should consider how existing users will migrate - hopefully there'll be no impact if they re-deploy to Heroku, but are there other providers or situations that we should document in the release notes, perhaps?

Also, we could just make the leap to 2.3.x directly - I'm not aware of any breaking changes, but we'd have to change the "minimum requirements" in the README and there aren't any important improvements in 2.3.x that we can't live without, so probably no need yet.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: ruby
 rvm:
-  - 2.2.1
+  - 2.2.5
 
 services:
   - postgresql
--- a/Gemfile
+++ b/Gemfile
@@ -1,5 +1,7 @@
 source 'https://rubygems.org'
 
+ruby "2.2.5"
+
 gem 'rails', '4.2.7'
 
 # Use postgresql as the database for Active Record
@@ -158,5 +160,3 @@ group :production do
   gem 'rails_12factor'
   gem 'unicorn'
 end
-
-ruby "2.2.1"


olliebennett:
Please add to [Rails 5 Update](https://github.com/helpyio/helpy/milestone/1) milestone.

arrowcircle:
I think ruby version should be removed from Gemfile. It should be managed by ruby version managers live rvm or rbenv. Otherwise, its impossible to use any other version and another point of maintenance of code.

hadifarnoud:
Ruby version should be in Gemfile. Both Flynn and Heroku setup the latest version if it isn't. which could be an issue in future

olliebennett:
Hey Oleg, I understand what you mean, and agree it seems arbitrary to "lock" the Ruby version.

The logic is that we require a minimum Ruby version (e.g. `2.2.2` beyond `Rails 5`), and need to convey this dependency, ideally with no user input.

I have noticed that the following is actually valid in the `Gemfile`, although I've never seen it used before (no idea why not)<sup>+</sup>:

```ruby
ruby ">= 2.2.2", "< 2.4.0"
```

This - as you would expect - allows us to specify a valid range of Ruby versions, and also addresses @hadifarnoud's concerns about unexpected future changes (eg. in Ruby 2.4 or above).

On a side note, it's still advisable to use `rbenv` (or `rvm`) to manage Ruby versions and install the (required) version.

<sup>+</sup> I haven't checked whether this `Gemfile` syntax is supported on Heroku (for example) yet.

arrowcircle:
@hadifarnoud @olliebennett Hey! Thanks for the answers.
Is Heroku only one real target?

connorshea:
https://devcenter.heroku.com/articles/ruby-support#libraries

Here's the Bundler version on Heroku right now, I'm surprised it hasn't been updated in the last two months.

connorshea:
Ruby version specifiers now supported
https://devcenter.heroku.com/changelog-items/1014

azul:
#339 seems related

olliebennett:
Great, let's close this as the linked issue is indeed related, and more up to date.

PR #253, opened 2016-07-22
JSHint: Resolve violations and run on TravisCI

olliebennett:
I noticed [JSHint](http://jshint.com/) isn't being run on the app's JavaScript code. The tool really helps keep things clean and error-free, as well as enforcing some style conventions (such as use of semi-colon).

All rules can be disabled (globally, or per-file) if required, but I've stuck with the default configuration for now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,8 @@ services:
 before_script:
   # Setup Test Database
   - psql -c 'create database helpy_test;' -U postgres
+  # Prepare 'jshint' executable for JavaScript linting
+  - npm install jshint
 
 script:
   # Load database schema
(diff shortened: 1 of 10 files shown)


olliebennett:
This one's good to go if you're happy, @scott :)

The only non-obvious change is `$(this).val() === ""` (was just `==`) but the behaviour will be unchanged whether `val()` returns `null`, `undefined` or an empty string:

```javascript
undefined == '' // false
undefined === '' // false

null == '' // false
null === '' // false

"" == "" // true
"" === "" // true
```

scott:
Yeah that probably should have been `===` :)  Nice refactoring!

PR #254, opened 2016-07-22
Support for multiple attachments

scott:


olliebennett:
Nice one - cool feature :) Will test when I get a chance

scott:
You can see I messed it up last week but this one actually works.  There is still one issue I discovered if you send an email with just attachments and no message it does not record them.  I think thats an edge case but still should be resolved.

PR #257, opened 2016-07-25
Improve update topic#topics controller

shaktik1989:


scott:
Awesome work @shaktik1989 Thanks again for this!

PR #261, opened 2016-07-27
update Persian translation

hadifarnoud:
fixed a few typo
translated new strings


hadifarnoud:
can you please merge this one?

scott:
Yes.  You should consider submitting your translations to the `devise_invitable` gem as a PR as well, since they don't have Persian.

hadifarnoud:
will do @scott. can you please add translation function to email views as well? or should I just create templates for Persian?

scott:
Which emails are you referring to specifically?  The main helpy email messages are already translated via the standard `.yml` files.

hadifarnoud:
ticket reply emails etc.

they are still in English even though I changed default locale to Persian. 
screen shot 1395-05-11 at 9 43 39 am

We have to do RTL thing for email templates as well. should we just have a different templates for RTL languages?

scott:
Hmmm.. yes it seems that the HTML version in particular will need a class or something to force it to RTL.  As for the translation, I am not sure why that is not working, let me take a look and see if I can see why not.

hadifarnoud:
any update on this? should I open an issue?

scott:
Yes that would be helpful.

PR #265, opened 2016-08-02
Add agent notications

scott:
Adds email notifications when a new private discussion or reply is posted.


test/models/post_test.rb
```diff
@@ -138,4 +138,72 @@ class PostTest < ActiveSupport::TestCase
     assert_equal(1, Topic.find(4).assigned_user_id, "Topic should be assigned to user 1")
 
   end
+
+  # Notifications Specs
+
+  test "Should send an admin notification of a new private topic created, if enabled" do
+    User.first.settings.notify_on_private = "1"
+
+    assert_difference('ActionMailer::Base.deliveries.size', 1) do
```
scott: @olliebennett Do you know if `ActionMailer::Base.deliveries.size` is available in the context of a model test?  This functionality works when I user test it, but the test fails.  Even if I remove the conditional sending logic out of the model.

PR #243, opened 2016-07-18
[WIP] Consolidate Internationalization 

olliebennett:
I thought it might be useful to consolidate issues like XYZ and XYZ by replacing them with a tool to automate the job somewhat. I've found the [i18n-tasks gem](https://github.com/glebm/i18n-tasks) which offers `i18n-tasks unused` and `i18n-tasks missing`, both of which help identify problematic translations.

Do you think it'd be helpful to auto-translate these missing strings using Google Translate? (the gem also automates this). I would be happy to provide a PR if so, and look through the missing/unused translation strings to do some cleanup...

An alternative option is to use a service like [Transifex](https://www.transifex.com/) to host and remotely manage the translations, but that may be overkill.

A small but representative sample of the output of `i18n-tasks missing` is shown below, to give an idea of the "problems" in the current code:

```
➜  helpy git:(feature/i18n-checks) ✗ i18n-tasks missing
warning: parser/current is loading parser/ruby22, which recognizes
warning: 2.2.4-compliant syntax, but you are running 2.2.1.
warning: please see https://github.com/whitequark/parser#compatibility-with-ruby-mri.
i18n-tasks: [WARN] 'pt_BR.results_found' was a leaf, now has children (value <- scope conflict)
...
i18n-tasks: [WARN] 'zh_TW.upvote' was a leaf, now has children (value <- scope conflict)
Missing translations (725) | i18n-tasks v0.9.5
+-------------------------------------------------+----------------------------------------------------------+-----------------------------------------------------------------------------------------------------------------------------------------------+
| Locale                                          | Key                                                      | Value in other locales or source                                                                                                              |
+-------------------------------------------------+----------------------------------------------------------+-----------------------------------------------------------------------------------------------------------------------------------------------+
|                       ca                        | activerecord.helpers.submit.post.create                  | en                                                                                                                                            |
|                       ca                        | design                                                   | en Design                                                                                                                                     |
|                       ca                        | edit_user                                                | en Edit User                                                                                                                                  |
|                       ca                        | general                                                  | en General                                                                                                                                    |
|                       ca                        | international                                            | en International                                                                                                                              |
|                       ca                        | mark_as_resolved                                         | en Mark as resolved                                                                                                                           |
|                       ca                        | no_helped                                                | en We're sorry this didn't help you.  Please contact us for more help!                                                                        |
|                       ca                        | settings                                                 | en Settings                                                                                                                                   |
|                       ca                        | widget                                                   | en Embeddable Widget                                                                                                                          |
[...]

Issue #61, opened 2016-02-29
Missing i18n: Admin Views


scott:
Admin Topic:

- Move forum dropdown: The word "Move"
- Post Dropdown: Not translated at all
- Reply to this Topic

Admin creating a new topic:

- Discussion Management header

Article Edit:

- "Additional Settings" Header
- Article Status: Published vs. Draft
- Save Changes

Admin Communities

- Admin Communities Header
- Community TH

Issue #40, opened 2016-02-22
Missing i18n keys


scott:
there still seem to be some translated strings that aren't displayed properly. 
URL: .../topics/new

ask a question: "subject"
URL: .../users/edit
my profile: "signature"
URL: .../community/.../edit 
 Allow topic voting
 Allow post voting
URL: .../ticket/...
Ticket Number:
Opened On:

PR #273, opened 2016-08-05
Updates to translations, add missing keys

scott:


scott:
Related to #40, #61, #243

Issue #276, opened 2016-08-08
Add ability to backup to CSV/import from CSV for backing up an entire Helpy


scott:
This feature will allow admin level users to create and download a full backup of their data in CSV form.  The UI should present an interface for choosing to backup just categories/docs, or include users/forums/topics/posts as well.  There should also be an import/restore function that lets admins pick a file to restore it.

We will add more requirements here as time goes on, but to start:
- [ ] Create UI for selecting scope of backup
- [ ] Select gems for assisting with export/import
- [ ] The CSV should have a version stamp/ datestamp

scott:
This branch may be helpful- https://github.com/bolser/helpy/tree/dev/helpjuice-importer

scott:
@shaktik1989 here is documentation for the import function.  Let me know what questions you have:

**Import Requirements**

The import feature of Helpy should allow users to restore a Helpy exported backup file, or bulk load new records by formatting a CSV to match that supported by Helpy.

There are two general cases when an import would be used:

1. To restore a backup CSV
2. To import data from another system

The import system will handle both of these although #2 would expect that the user should format their data in a way that matches the Helpy export template and configure associations on their own by manually populating primary and foreign keys to match.

**User Stories/Requirements**

1. All import functions should be constrained to admin level users.

2. The UI should allow for uploading CSV files.  The file select should be constrained to .csv only.

3. Similar to the way exports work, a new import job record should be saved with a pointer to the CSV file uploaded.  A worker will handle the actual processing of the CSV in the background.

4. The import job table should also track the status of the job, to provide a UI that shows the progress of the import. 
  a. Statuses are [“In Progress”,”Completed”,”Error”].
  b. There should be a notes section for storing additional info about the job.

5. Before beginning processing the CSV, the column mapping of the CSV should be validated.  
  a. If the column mapping does not match that which is expected, the status will be set to “Error” and the job will quit.  
  b. The notes for that job will be set to “The import CSV file was incorrectly formatted.”

6. As the import worker processes the csv, it should check each record ID to determine how to handle it.  
  a. If there is an ID and it finds a duplicate, it will skip that record. Notes for that job will include the ID of the record skipped.
  b. If there is an ID  and no duplicate ID, the record will be inserted.
  c. If there is no ID present in the CSV, the record should be inserted and get an ID assigned by the database.

7. Once the import has completed, add a note “Import completed %datetime%”

8. On completion of the import, send an email to admins indicating the job has completed and including any notes added to the job record.

[...]

PR #284, opened 2016-08-11
export data into csv

shaktik1989:


scott:
Looks like a great start @shaktik1989 What I envisioned was to generate these CSV files as a background worker since in production they could be quite large.  Once the csv is ready, the admin should be notified by email with a protected link or something like that.  

#276

shaktik1989:
@scott I have update the

scott:
Cool I will take a look!  By the way, Rails updated to 4.2.7.1, and if you pull from `master` your integration tests should pass again (unless there is another issue.)

app/jobs/import_job.rb
```diff
@@ -0,0 +1,65 @@
...
+    end  
+  end
+
+  def user_import(obj_hash)
+    object = User.new
+    object.attributes = obj_hash
+    object.password = User.create_password
+    object.sign_in_count = 0
+    object.save!
+  end
```
scott: Regarding your comment about errors, we probably should add some kind of rescue or error handling here.  On one hand this could mean one or more of:
- storing the full hash into a text_field for inspection
- attempting to identify the specific problem and recording it (not sure how feasible this is with mass assignment)
- just report a count of the number of rows that failed.
- other ideas?

app/jobs/import_job.rb
```diff
@@ -0,0 +1,65 @@
...
+        end
+      end
+    else
+      #notes The import CSV file was incorrectly formatted.
+    end  
+  end
+
+  def user_import(obj_hash)
+    object = User.new
+    object.attributes = obj_hash
```
scott: Are validations called here?

scott:
This looks great so far! Overall it seems to work pretty well, but we need to suppress notifications and emails when doing an import of topics and posts.  Also, on the import page, it would be much nicer if we used something like http://www.dropzonejs.com/.  And had a UI that showed the pending uploads and their status- I think you just haven't gotten to that yet.

As far as error checking, I think we need some test cases to describe and verify that its working.  I uploaded a couple invalid user rows and got an email report saying "no problems".  Is this the expected behavior?

PR #308, opened 2016-08-27
Fixes #305

### LINKED ISSUE #305: Common reply parsing fail
Common reply text stops at '.
Example:
"You've been banned."
Becomes:
"You"

No client or server side errors.
It appears to be saved correctly on the server but when I reply to some discussion and choose a common reply, I'll get this bug.

CGA1123:
The use of single quotes meant that common replies including the `'` character would not be correctly inputted as an `option`'s value, this fixes that.


CGA1123:
Previous HTML output
```
<option value="Using single quotes " does="" not="" work'="">Using unescaped single quotes ' results in truncation</option>
```
Now:
```
<option value="Using single quotes ' works">Using unescaped single quotes ' results in truncation</option>
```

Issue #304, opened 2016-08-20
Refactor and write tests for insights statistics


scott:
The insights view has a lot of great statistics, but needs some cleanup.  A lot of the queries are inline and could be moved off to the `Topic` class.  This would make it easier to write tests for them as well.

PR #313, opened 2016-08-30
Cleaning up Admin::DashboardController

CGA1123:
I noticed that some views within the project include code that would ideally be within controllers.


CGA1123:
the `_agent_stats.html.erb` partial also has a couple statements executing queries from the view, which is not ideal.

scott:
#304

app/controllers/admin/dashboard_controller.rb
```diff
@@ -46,12 +31,28 @@ def stats
     @responded_topics = Topic.where(id: responded_topic_ids)
     @closed_topic_count = @topics.closed.count
 
-    @posts = Post.where('created_at >= ? AND created_at <= ?', @start_date, @end_date)
+    @posts = Post.where(created_at: @start_date..@end_date)
```
scott: Nice, didn't know you could write it that way

CGA1123:
For moving to code from `_agent_stats.html.erb` to the controller, the only idea I have so far is to create a few hashes, one for each variable that is currently declared in the view, using `agent.assigned_user_id` as the key, and then pull from the hash within the view. 

I don't know whether this is the best solution though.

CGA1123:
Also, I'm not sure that the calculation for delays works properly, what happens if the second post on a topic is not from an agent?

`delays = @responded_topics.map { |t| t.posts.second.created_at - t.created_at }`

scott:
I think it would be ideal to move code from `agent_stats` to the model where possible because it would make it cleaner and easier to test, and I just think thats where calculations like `responded_topic_count` belong.

As for the delays, you may be right.  Perhaps adding a scope on `Post` for agent would help?

CGA1123:
To keep consistent, would some code currently in `Admin::DashboardController` be better off in the model as well?

For example `@reponded_topics`, `responded_topic_ids`, and `@agents`

app/controllers/admin/dashboard_controller.rb
```diff
@@ -44,15 +47,11 @@ def stats
...
-
-    # responded_topic_count = @responded_topics.where(assigned_user: agent.assigned_user).length
-    # closed_topic_count = Topic.undeleted.where(assigned_user: agent.assigned_user).closed.count
-    # posts_count = @posts.where(user: agent.assigned_user, kind: 'reply').count
-    # delays = @responded_topics.where(assigned_user: agent.assigned_user).map { |t| t.posts.second.created_at - t.created_at }
+    @agents_topic_count =           @agents.each.map { |agent|  [agent.id, @topics.where(assigned_user: agent).count] }.to_h
+    @agents_responded_topic_count = @agents.each.map { |agent|  [agent.id, @responded_topics.where(assigned_user: agent).count] }.to_h
+    @agents_closed_topic_count =    @agents.each.map { |agent|  [agent.id, Topic.undeleted.where(assigned_user: agent).closed.count] }.to_h
+    @agents_post_count =            @agents.each.map { |agent|  [agent.id, @posts.where(user: agent, kind: 'reply').count] }.to_h
+    @agents_delays =                @agents.each.map { |agent|  [agent.id, median(@responded_topics.where(assigned_user: agent).map { |t| t.posts.second.created_at - t.created_at })] }.to_h
```
CGA1123: Working on moving these assignments into a single iteration.

CGA1123:
Only problem with moving some of this code into the model is the dependence on `@topics`

scott:
What do you think about testing this? Could you write some tests?

CGA1123:
I'm more familiar with rspec than minitest but I can give it a shot.

scott:
Awesome work! I am really hoping we can add some tests to validate that the stats are showing what they should.  Do you have any ideas for this?

CGA1123:
Thanks!

Yeah I suppose it's pretty straightforward, just involves a lot of setting up for the tests.

CGA1123:
Can anyone point me in the right direction for testing whether the controller instance variables are being set correctly in minitest.

Maybe the equivalent for `assigns['instance_variable']` in rspec?

scott:
You can access instance variables using `assigns(:name_of_variable)`  Examples here: http://api.rubyonrails.org/v4.2/classes/ActionController/TestCase.html

Also, this has some conflicts after I merged #319 from @Rynaro which touched some of the same files.

CGA1123:
@scott Thanks! I'll have a look and try to get those test written asap.

I fixed the conflicts.

CGA1123:
@scott What do you think about changing `date_from_params` to use `params[:label]` to set the start and end dates, rather than having the `start_date` and `end_date` params in the url?

CGA1123:
@scott I'm not great with fixtures, so writing test for the DB queries made by the controller is pretty difficult for me.

However, I've added tests for other functionality.

Issue #301, opened 2016-08-16
Toggle Tickets & Forums off (settings.yml)


brianjking:
Hi,

I see that `settings.yml` has the following feature toggles included in it. I set tickets & forums to be false and deployed to Heroku and these features are still available.

Based on this I'm guessing it's not yet fully implemented. When I checked the roadmap on Trello, I couldn't find any mention of when this is planned for.

Please advise. Thanks!

````
# Toggle functionality (not fully implemented)
tickets: true
forums: true
knowledgebase: true
````
https://github.com/helpyio/helpy/blob/master/config/settings.yml#L4-L7

brianjking:
@scott - Sorry to bug you, just wanted to see if you could quickly enlighten me here? Thanks for all you do. :+1:

scott:
Hey @brianjking I actually replied to your comment over here [http://support.helpy.io/en/knowledgebase/18-configuring-your-helpy/docs/22-configuring-your-helpy-settings](http://support.helpy.io/en/knowledgebase/18-configuring-your-helpy/docs/22-configuring-your-helpy-settings) but the bottom line is that those capabilities are not enabled yet.

brianjking:
@scott Thanks, sorry, didn't see that comment. Is this feature on the roadmap anywhere so that I can track it?

scott:
Are you hoping to use it?

brianjking:
Yeah, I really just want to use the knowledge base feature if possible.

brianjking:
@scott - Anyway I can help test this or anything else I can do? Thanks!

hadifarnoud:
I would love this feature too. we prefer to turn forum off.

scott:
Okay, it seems like a pretty common request!

brianjking:
@scott Thank you! I have a question though, even though I logged in and disabled the functionality outside of the knowledgebase there's still a "sign up" option.

What are my options for disabling the sign-up option for end users?

As a follow-up what do I need to enable settings wise to get an email/push notification/admin notification/smoke signal for when people click "yes/no" to an article being helpful?

scott:
Yeah I should probably remove the signup option.  Good catch... The yes/no will automatically fire an event into Google Analytics if you have provided that ID.  Then you can use a conversion event to quickly see the % approval rate for each article.

brianjking:
@scott I've created a new request to track the sign-up disable option in #334.

Thanks for getting this issue closed (#301).

PR #325, opened 2016-09-16
Implement optional features for tickets, forums, knowledgebase

scott:
Full implementation of feature toggles for knowledgebase, community forums, and private tickets.


scott:
#301

Issue #66, opened 2016-03-03
Missing breadcrumb in mobile locale selection page


scott:
The translation for the root breadcrumb is missing.  Also, the link to go to this page should be hidden if there are no alternate translations.

![screen shot 2016-03-02 at 7 39 36 pm](https://cloud.githubusercontent.com/assets/14369/13483064/8b2dbade-e0ae-11e5-8336-9424c5aac8f7.png)

olliebennett:
As a quicker way to identify this more generally, how about forcing exceptions to be raised on missing translations?

Basically, we just need to uncomment the following in `development.rb` and `test.rb`:

```ruby
# Raises error for missing translations
config.action_view.raise_on_missing_translations = true
```

More info [here](https://robots.thoughtbot.com/foolproof-i18n-setup-in-rails).

PR #331, opened 2016-09-19
Add missing translations in breadcrumbs

lidimayra:
In order to correct bug #66


scott:
Hi @lidimayra this looks like a great PR.  Thanks!

Issue #323, opened 2016-09-15
Add better CSS handling for printing of tickets, other views


scott:
Create a stylesheet for printing.

hin101:
I can have a look at this one for you.

hin101:
#341 removed href links from print view

scott:
Can you make one more enhancement to this and remove the "support pipeline" from the printable area?  I don't think this adds much value if someone is trying to print a ticket.

hin101:
Issue fixed in #352

PR #333, opened 2016-09-20
Fix css for printing

hin101:
Added printer friendly versions of various pages.

Reference: Issue #323


scott:
This is a huge improvement 👍 I still noticed some differences in the way printing is handled depending on the browser.

PR #338, opened 2016-09-23
Enhance support for local SMTP servers

NoyaInRain:
A common way for sending emails is to use a local SMTP server, most often not requiring
authentication and TLS and providing TLS only with a self-signed certificate (this is, for example,
the default configuration for Postfix on Debian/Ubuntu). Thus, skip authentication if no SMTP
credentials are set and disable TLS if the SMTP address is set to localhost. Along the way remove
the obsolete mail initializer.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -110,11 +110,12 @@ class ApplicationController < ActionController::Base
     Griddler.configuration.email_service = AppSettings["email.mail_service"].present? ? AppSettings["email.mail_service"].to_sym : :sendgrid
 
     ActionMailer::Base.smtp_settings = {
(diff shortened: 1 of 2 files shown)


NoyaInRain:
Dear everyone, thanks for your awesome work on Helpy, I really like it so far. Here is a small patch that makes it easier to set up Helpy on a standard Linux box. I'm not a Ruby dev, and this my first attempt at writing some Ruby, just so you know ;) .

CGA1123:
Looks Good @NoyaInRain!

However I think it would be better to use the newer `symbol_key: value` hash syntax, rather than. the older `:symbol_key => value`

scott:
Yeah I agree @CGA1123

PR #339, opened 2016-09-23
Relax required Ruby version

NoyaInRain:
If an exact Ruby version is specified in the Gemfile, bundler will abort with an error if a newer,
yet compatible, Ruby interpreter is used. Thus, only specify the minimum required version,
simplifying the setup of Helpy on systems where a current Ruby interpreter is available.

Using Ruby version operators is only supported by newer Bundler versions, so make the current one
available in the Travis CI environment.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,14 @@
 language: ruby
 rvm:
-  - 2.2.1
+  - 2.2.5
+  - 2.3.1
 
 services:
   - postgresql
+
+before_install:
+  # Install a current Bundler version
+  - gem install bundler
 
 before_script:
   # Setup Test Database
--- a/Gemfile
+++ b/Gemfile
@@ -181,4 +181,4 @@ group :production do
   gem 'unicorn'
 end
 
-ruby "2.2.1"
+ruby '>= 2.2', '< 2.4'


NoyaInRain:
I know in the Ruby community it is common to use RVM, rbenv or the like to install multiple Ruby versions, but relaxing the required version to only specify a minimum will remove one hassle to set up Helpy. For example, on our servers, we use one current version of Ruby to run multiple Ruby applications.

Rynaro:
I really like this idea of relax Ruby version management. I my common development environment uses Ruby 2.3.0, and Helpy works as well, without any issues. But, if I need to commit any improve or fix here, I simply unmap the Gemfile changes. Or if I need to commit the Gemfile, I revert the Ruby version.
But I don't see any problem in Ruby version flexibilize. But this flex permissions needs a minimal control to avoid the Ruby unstable versions. 
What do you think about this @scott ?

CGA1123:
If the ruby version were to be relaxed, should we not add more ruby versions to the `.travis.yml` so they can also be tested?

NoyaInRain:
Dear @Rynaro and @CGA1123 , thanks for your feedback. I've had a look at the [Ruby version policy](https://www.ruby-lang.org/en/news/2013/12/21/ruby-version-policy-changes-with-2-1-0/) and to my surprise minor releases may be API incompatible. Thus, I've set the upper bound to `< 2.4`, which should also serve as

> a minimal control to avoid the Ruby unstable versions

Additionally I've added the latest tiny/patch version of each compatible Ruby release to `.travis.yml`, as suggested.

CGA1123:
LGTM! 👍

scott:
Do you know if this will affect Heroku compatibility?  It looking like Heroku is going to move to a branch anyways because of some other things I am about to merge in like Carrierwave which is not very compatible with Heroku out of the box.

NoyaInRain:
I moved away from Heroku (to self-hosting) last year, so I have no way to test and confirm or deny ;) . Looking through the Heroku docs, the Ruby version should be fine (current and older Rubies are supported), but the Bundler version could indeed proof problematic:

* For **now**, [Bundler is locked to 1.11.2](https://devcenter.heroku.com/articles/ruby-support#libraries)
* But it seems **soon** [a current Bundler version will be available](https://github.com/heroku/heroku-buildpack-ruby/blob/master/CHANGELOG.md)

connorshea:
Ruby version specifiers now supported
https://devcenter.heroku.com/changelog-items/1014

azul:
@scott looks like this can be merged then, right?

scott:
Yeah we have deprecated the one-click install to Heroku so removing the Ruby specification will not break it, which was the hold up before.

noyainrain:
Hey @scott , I've noticed you have merged (thanks :+1: ), but then reverted the patch. I've had another look at the problem you mention:

> Broken Travis with "no env variable set"

*no env variable set* is just a message Travis prints if no environment variables are set in the config file. Looking closer at the [build job](https://travis-ci.org/helpyio/helpy/jobs/191986401), it erred because of 6 failing test cases:

```
1) Failure:

EmailProcessorTest#test_an_email_to_the_support_address_with_email_with_numbers_should_create_a_new_user_and_topic_with_status_new [/home/travis/build/helpyio/helpy/test/lib/email_processor_test.rb:34]:
"ActionMailer::Base.deliveries.size" didn't change by 2.
Expected: 61
Actual: 59

[...]
```

These problems, however, seem to have been fixed by now. I've rebased the branch on current master and all unit tests pass, locally and [on Travis](https://travis-ci.org/helpyio/helpy/builds/195924650).

Thus, I've opened a follow-up pull request #531, ready for merging :) .

PR #341, opened 2016-09-24
remove href links from print view

hin101:


hin101:
In relation to #323

CGA1123:
@hin101 `scss-lint` doesn't like the use of `!important`

hin101:
removed !important in latest commit

PR #342, opened 2016-09-24
Notification Mailers not scoped correctly, fixes #335

### LINKED ISSUE #335: Notification emails are sent out to non-agents
Still trying to narrow this down, but:

When receiving new topics/tickets via email, notification emails should be sent out to those agents who have enabled email notifications. Currently an email is generated and sent out to some other registered users as well - even some who haven't activated their accounts (another user having sent in an email but never set a password/used the web interface). The log from helpy states that the email was sent to "agents", in Sendgrid's log there are 10-ish recipients per email. We have 5 agents registered, and more than 10 users in total.

I'll try to understand which users receive emails and why as soon as I've set up a new test environment that looks more like production =)

scott:


scott:
#335

Issue #343, opened 2016-09-25
Flag for Review Feature


scott:
# Overview

The community feature of Helpy allows users to create and reply to messages in an open forum and interact with each other through discussion threads.  The post flagging feature makes it possible for other users to alert forum administrators that possibly objectionable content has been created in the community, and should trigger a review by the administrators.
# Stories
1. Each post displayed in the community forums should have a tiny link that says "Flag for review".
2. The "Flag for Review" option should only work for users who are currently logged in.  If a user is not logged in, clicking on the link will prompt the user to log in.
3. When a user clicks "Flag for Review" the system should automatically create a private discussion topic that is shown as being from the user who has reported the thread. 
   a. After clicking, a modal should be shown that asks the user to add a reason for flagging the post.  This input should be included in the private discussion that is being created.
   b. The title of this discussion should be "Flagged for review: [name of the parent topic of the post that was flagged]"
   c. The body of this discussion should provide a link to the admin page for the flagged topic/post.

CGA1123:
Allowing the user to add a reason for reporting might be a nice touch as well.

hin101:
Hey Scott, I am working on this one for you

scott:
I think adding a reason for reporting input is a good idea, it will also reduce the likelihood of people accidently reporting posts for no reason.  I will update the stories.

hin101:
#349

hin101:
1st pull request: #349

PR #349, opened 2016-09-30
Flag for review feature

hin101:
In this first commit, I have added the frontend for the flag post for review feature.


hin101:
In relation to story/issue #343

app/views/layouts/_footer.html.erb
```diff
@@ -24,3 +24,18 @@
...
+        <button type="button" class="close" data-dismiss="modal" aria-label="Close"><span aria-hidden="true">&times;</span></button>
+        <h3 class="modal-title">Flag Post for Review</h3>
+      </div>
+
+      <div class="modal-body row">
+        <p>This modal allows you to flag a post for review</p>
+      </div>
+    </div><!-- /.modal-content -->
+  </div><!-- /.modal-dialog --> 
+</div><!-- /.modal -->
```
scott: I know this is a WIP so maybe you planned to change this, but I am wondering if you can find a way to re-use the modal construct for login, since this is not very DRY.  Also, it does not really need to appear on every page, just the topic detail.

hin101:
Functionality completed, need to write tests and validations

db/migrate/20161001090839_create_flags.rb
```diff
@@ -0,0 +1,11 @@
+class CreateFlags < ActiveRecord::Migration
+  def change
+    create_table :flags do |t|
+      t.integer :post_id
+      t.integer :generated_topic_id
+      t.text :reason
+
+      t.timestamps null: false
```
scott: I think timestamps should be included because it might let us use some kind of time decay counter in the future to automatically disable posts that have been flagged many times recently.

test/controllers/flag_controller_test.rb
```diff
@@ -0,0 +1,7 @@
+require 'test_helper'
+
+class FlagControllerTest < ActionController::TestCase
+  # test "the truth" do
+  #   assert true
+  # end
+end
```
scott: We should add some basic test coverage here.  I have used the `shoulda` gem which has some nice helpers for testing relationships- see the other models.

app/views/posts/index.html.erb
```diff
@@ -48,3 +48,23 @@
...
+  <div class="modal-dialog">
+    <div class="modal-content">
+      <div class="modal-header">
+        <button type="button" class="close" data-dismiss="modal" aria-label="Close"><span aria-hidden="true">&times;</span></button>
+        <h3 class="modal-title">Flag Post for Review</h3>
+      </div>
+
+      <div class="modal-body row">
+        <div class="col-md-12">
+        <%= bootstrap_form_for :flag, :url => topic_flags_path(@topic) do |f| -%>
```
scott: This is passing the `@topic` instead of the `@post`.  As a result, the record does not seem to be saved in the `Flags` table.  It does generate a private topic though, but the reason is blank.

app/views/posts/index.html.erb
```diff
@@ -18,7 +18,7 @@
 		<div id="posts">
 			<%= render :partial => 'post', :collection => @posts %>
 		</div>
-
+        <%= link_to 'Flag for Review', '#', data: {toggle: "modal", target: user_signed_in? ? "#flag-modal" : "#login-modal"} %>
 		<%
```
scott: This should actually go into the partial since we are going to allow flagging of individual posts.

app/views/posts/index.html.erb
```diff
@@ -18,7 +18,7 @@
 		<div id="posts">
 			<%= render :partial => 'post', :collection => @posts %>
 		</div>
-
+        <%= link_to 'Flag for Review', '#', data: {toggle: "modal", target: user_signed_in? ? "#flag-modal" : "#login-modal"} %>
```
scott: This should actually go into the partial since we are going to allow flagging of individual posts.

app/controllers/flags_controller.rb
```diff
@@ -0,0 +1,36 @@
...
+
+    @topics = @forum.topics.new(
+    	name: "Flagged for review: #{@topic.name}",
+    	private: true)
+
+    @user = current_user
+    @topics.user_id = @user.id
+
+    if @topics.save
+      @flag = Flag.new(
```
scott: I am thinking this logic might be reversed, so it only generates the topic if the flag is saved.  Then you could actually move the topic generation into the Flag model- making the code more modular and the controller thinner.  Thoughts?

config/routes.rb
```diff
@@ -47,7 +47,8 @@
       resources :topics
     end
     resources :topics do
-      resources :posts
+      resources :posts 
+      resources :flags, only: [:create]
```
scott: I think you want to have flags nested under posts instead. You can put it either inside the topics block or under posts on line 53.

Issue #346, opened 2016-09-28
Refactor CSS for Themes


scott:
The themes that are included with Helpy (Flat and Light) both use a modified copy of the base.scss from core Helpy.  Instead of using this, the core should be maintained in base.scss, and any updated or overriding changes should be kept in the themes.

PR #351, opened 2016-10-01
Open attachments in new window, fixes #350

### LINKED ISSUE #350: Attachments should open in new window

scott:
Also introduces shared scss for shared/structural css.


scott:
See #346

PR #352, opened 2016-10-02
Fix css for printing - Remove support pipeline and user info

hin101:


app/assets/stylesheets/topics.scss
```diff
@@ -1,3 +1,13 @@
...
+
+@media print {
+  #pipeline-wrapper {
+    display: none;
+  }
+
+  #left-col-user-info {
+    display: none;
+  }
+}
```
CGA1123: @hin101 `scss-lint` is complaining about there not being a new line at the end of this file!

hin101: ill do that change and recommit

PR #357, opened 2016-10-03
Make text show rtl for emails

### LINKED ISSUE #272: emails are still in english even though locale is set to `fa`
ticket reply emails are still in English even though I changed default locale to Persian. 
screen shot 1395-05-11 at 9 43 39 am

CGA1123:
Closes #272

### DIFF
--- a/app/mailers/topic_mailer.rb
+++ b/app/mailers/topic_mailer.rb
@@ -3,7 +3,6 @@ class TopicMailer < ActionMailer::Base
 
   def new_ticket(topic)
     @topic = topic
-    @locale = @topic.locale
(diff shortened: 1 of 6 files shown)


hadifarnoud:
I can confirm that it works.

however there is a translation issue in new user email. the title says `welcome to`. seems like it's not going thru i18n.


[2.eml.zip](https://github.com/helpyio/helpy/files/508183/2.eml.zip)

CGA1123:
@hadifarnoud 

#358 should fix that issue.

CGA1123:
@scott Thanks!

PR #362, opened 2016-10-05
Welcome Email Toggle

### LINKED ISSUE #360: Toggle welcome message
Currently every new user gets a welcome message with a link to "claim" their account.  This really should be configurable:

Add a settings toggle to enable or disable the welcome new user message that gets sent out when a new user is added to the system. There should also be a way to customize the message- although this could get messy really quickly across all locales.

CGA1123:
Closes #360

### DIFF
--- a/app/mailers/user_mailer.rb
+++ b/app/mailers/user_mailer.rb
@@ -2,6 +2,7 @@ class UserMailer < ActionMailer::Base
   add_template_helper(ApplicationHelper)
 
   def new_user(user, token)
+    return unless (AppSettings['settings.welcome_email'] == "1" || AppSettings['settings.welcome_email'] == true)
     @user = user
     @token = token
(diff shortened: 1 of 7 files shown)


app/mailers/user_mailer.rb
```diff
@@ -2,6 +2,7 @@ class UserMailer < ActionMailer::Base
   add_template_helper(ApplicationHelper)
 
   def new_user(user, token)
+    return unless (AppSettings['settings.welcome_email'] == "1" || AppSettings['settings.welcome_email'] == true)
```
CGA1123: Does anyone know how I might access `ApplicationController#welcome_email?`, at the moment this is not very DRY.

scott:
Awesome! can you also add a test to assert that no mailer is sent out?

PR #363, opened 2016-10-06
 Add scalingo one click deploy

EtienneM:
Add a link to deploy in one click to the Scalingo platform. Scalingo is a PaaS that supports Rails pretty well.

These changes add the scalingo.json file to provide deployment information to Scalingo and add a _Deploy to Scalingo_ button on the README.md.

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,1 +1,2 @@
 web: bundle exec unicorn_rails -p $PORT -c ./config/unicorn.rb
+postdeploy: bundle exec rake db:setup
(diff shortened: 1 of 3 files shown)


scott:
Does Scalingo have a persistent file storage?  Since we have moved to a default option for file storage, we have deprecated the Heroku one click install since they do not.  Does Scalingo have a free plan like Heroku?

EtienneM:
OK I see. Scalingo, like other PaaS, do not provide a persistent file storage.

Scalingo offers a 30 days free trial but no free plan.

PR #365, opened 2016-10-07
[WIP] Adds polling for ticket management views

scott:
This PR adds a polling capability to the discussion listing and topic view.


app/controllers/admin/topics_controller.rb
```diff
@@ -27,7 +27,7 @@
 class Admin::TopicsController < Admin::BaseController
 
   before_action :verify_agent
-  before_action :fetch_counts, :only => ['index','show', 'update_topic', 'user_profile']
+  before_action :fetch_counts, :only => ['index','show', 'update_topic', 'user_profile', 'refresh_list', 'refresh_topic']
```
CGA1123: Should the new hash syntax not be preferred? `only: [:index, :show, :update_topic]`

app/controllers/admin/topics_controller.rb
```diff
@@ -305,6 +305,52 @@ def assign_team
     end
   end
 
+  def refresh_list
+#    @status = params[:status]
+#    get_tickets if params[:status].present?
+
+    @status = params[:status] || "pending"
+    if current_user.is_restricted? && teams?
+      topics_raw = Topic.all.tagged_with(current_user.team_list, :any => true)
```
CGA1123: Could `Topic.tagged_with` not be used her rather than `Topic.all.tagged_with`

app/controllers/admin/topics_controller.rb
```diff
@@ -305,6 +305,52 @@ def assign_team
...
+    else
+      topics_raw = Topic
+    end
+    topics_raw = topics_raw.includes(user: :avatar_files).chronologic
+    get_all_teams
```
CGA1123: could `get_all_teams` not be called in a `before_action`

app/controllers/admin/topics_controller.rb
scott: Hmm I am not sure, I just repeated how @jprberlin did it originally.  I will try it.

app/controllers/admin/topics_controller.rb
scott: Agreed.  Looks like I regressed :)

PR #368, opened 2016-10-08
hide unneeded admin elements when features are disabled

scott:
Hides some UI and navigation when forums, communities or kb are disabled

### DIFF
--- a/app/controllers/admin/dashboard_controller.rb
+++ b/app/controllers/admin/dashboard_controller.rb
@@ -9,9 +9,9 @@ class Admin::DashboardController < Admin::BaseController
   def index
     #@topics = Topic.mine(current_user.id).pending.page params[:page]
 
-    if current_user.is_admin? || current_user.is_agent?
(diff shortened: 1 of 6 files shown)


app/controllers/admin/dashboard_controller.rb
```diff
@@ -9,9 +9,9 @@ class Admin::DashboardController < Admin::BaseController
   def index
     #@topics = Topic.mine(current_user.id).pending.page params[:page]
 
-    if current_user.is_admin? || current_user.is_agent?
+    if (current_user.is_admin? or current_user.is_agent?) and (forums? or tickets?)
```
CGA1123: The [Ruby Style Guide](https://github.com/bbatsov/ruby-style-guide#no-and-or-or) recommends using `||` over `or`

CGA1123: I suppose it really is up to you, but it would be better to keep it consistent across the whole project.

#375 adds the cop that checks for this style violation, and also changes other occurrences of `and` or `or`.

PR #385, opened 2016-10-11
Fix dropdown menu bug

### LINKED ISSUE #373: Dropdown Menu Bug
The dropdown menu is covered by over posts.

Screenshot below:

9bc5de9af26d8f492ef5352b4b4e4445

Ideally the dropdown menu should show over the other elements.

CGA1123:
The bug was caused by the css classes `.media` & `.media-body`, as they define `overflow: hidden`.

~~This PR adds an 'override' file in `app/assets/stylesheets/overrides.scss` that is imported last in the pipeline, allowing css rules contained in there to override any previous rules. 
(I don't know whether this is the best approach, overriding `overflow: hidden` might break some
other stuff)~~

Fixes #373

### DIFF
--- a/app/assets/stylesheets/overrides.scss
+++ b/app/assets/stylesheets/overrides.scss
@@ -1,6 +1,0 @@
-// Any SCSS in this file will be loaded last in the pipeline
-.media,
-.media-body {
-  overflow: visible;
-  zoom: 1;
-}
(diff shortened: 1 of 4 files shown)


scott:
Aren't those core Bootstrap classes?

app/assets/stylesheets/overrides.scss
```diff
@@ -0,0 +1,6 @@
+// Any SCSS in this file will be loaded last in the pipeline
+.media,
+.media-body {
+  overflow: visible;
+  zoom: 1;
+}
```
scott: Can we do this in a more specific way to reduce the likelihood of unexpected side effects

Issue #390, opened 2016-10-13
Add additional date options and a freeform datepicker to insights


scott:
The current implementation of the insights daterange picker is extremely limited.  This enhancement will increase the flexibility of this by adding a start and end daterange picker, and two new predefined ranges- "last week" and "last month".

Here is a rough idea of how it will look:

![screen shot 2016-10-13 at 12 18 12 pm](https://cloud.githubusercontent.com/assets/14369/19361412/d2deb990-913f-11e6-968a-8487e41bc513.png)

The insights report is already set up to take a date range as params, so it should be pretty easy to add these, although watch for turbolinks problems with the calendar form (I ran into this with the similar form on dropdown for adding groups to tickets).

PR #392, opened 2016-10-16
Fix/390

SundayAdefila:
This PR implements issue #390 by adding the two default date ranges as well as adding a form for picking custom start and end dates. Implementation is as image attached.

additional date ranges

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -95,6 +95,7 @@ gem 'twitter-bootstrap-rails-confirm'
 gem 'rdiscount'
 gem 'selectize-rails'
 gem "bootstrap-switch-rails"
+gem 'bootstrap-datepicker-rails'
 
 gem 'config', '~> 1.1.0', git: 'https://github.com/railsconfig/config.git'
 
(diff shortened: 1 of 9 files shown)


scott:
Thanks Sunday!  I will get some feedback together for you.

scott:
So I was thinking about the way the date range filter looks and works, and think we would be better  to use this: https://uxsolutions.github.io/bootstrap-datepicker/?markup=component&format=&weekStart=&startDate=&endDate=&startView=0&minViewMode=0&maxViewMode=4&todayBtn=false&clearBtn=false&language=en&orientation=auto&multidate=&multidateSeparator=&keyboardNavigation=on&forceParse=on#sandbox

Can you also left align the start and end labels on with the same left margin that the other options have?

SundayAdefila:
Okay, looking into that now.

SundayAdefila:
this has been updated now, and rendering as follow: 


screen shot 2016-10-21 at 3 35 04 pm

screen shot 2016-10-21 at 3 01 03 pm

PR #394, opened 2016-10-17
Move upcase to css

hin101:
I have moved some cases from ruby to CSS.


olliebennett:
Excellent @hin101; looks good to me!

scott:
Yes, this looks great!

PR #397, opened 2016-10-18
Hungarian translation.

tcsizmadia:
Greetings,

I have created the Hungarian 🇭🇺 translation of Helpy, but did not add the `hu` locale to `config.i18n.available_locales` in `production.rb` but in `development.rb`.
I was not sure about that it would be appropriate.

Many thanks,
Tamas.


scott:
Thanks so much for doing this!  Awesome!

PR #420, opened 2016-10-28
Uploadable logo

scott:
Adds upload-ability in the design settings for changing the logo and favico


app/controllers/admin/settings_controller.rb
```diff
@@ -37,6 +37,11 @@ def update_settings
       AppSettings[setting[0]] = params[setting[0].to_sym]
     end
 
+    @logo = Logo.new
+    @logo.file = params['uploader.design.header_logo']
+    # binding.pry
```
muescha: debug commands should be removed

scott: Yeah apparently that one snuck in :)

PR #422, opened 2016-10-29
Fix issue 396 - Reject quotes from user name

### LINKED ISSUE #396: Usernames are being saved with quotes
Some email providers put the users name in quotes, so when a user is added through `email_processor` the user ends up getting saved with a name that includes quotes.  These should be cleaned prior to saving.

lccezinha:
Fix #396

Add a constant to make possible add another invalid chars for use name, and use a callback to verify if user name needs to be sanitazed before create a user.


app/models/user.rb
```diff
@@ -94,6 +96,7 @@ class User < ActiveRecord::Base
 
   after_invitation_accepted :set_role_on_invitation_accept
   after_create :enable_notifications_for_agents
+  before_create :reject_invalid_characters_from_name
   acts_as_taggable_on :teams
```
scott: Could you change to before_save so people can't go in and add invalid characters later on?  I doubt they would, but right now they could

lccezinha:
Hello @scott I will fix merge error, and I'll accept you suggestion to change callback and push it again soon.

lccezinha:
Hello @scott can you see my changes again ?

scott:
Sorry for the delay, I have been on vacation out of the country.  I think this looks good, let me double check it and merge it in

lccezinha:
Ok @scott, Thanks!

PR #423, opened 2016-10-30
Make topic subject editable

hin101:
Functionality done just need to style button and add some tests

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -46,11 +46,6 @@
 // Jtruncate plugin, http://www.jeremymartin.name/projects.php?project=jTruncate
 // modified by Scott Miller- remove animation, newline for more link
 
-$(document).ready(function() {
-  /* Activating Best In Place */
-  jQuery(".best_in_place").best_in_place();
-});
-
 (function($){
   $.fn.jTruncate = function(opts) {
     var defaults = {
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -41,6 +41,7 @@
 //= require selectize
 //= require bootstrap-switch
 //= require bootstrap-datepicker
+//= require rest_in_place
 
 // Jtruncate plugin, http://www.jeremymartin.name/projects.php?project=jTruncate
 // modified by Scott Miller- remove animation, newline for more link
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -11,6 +11,7 @@
 // about supported directives.
 //
 //= require jquery
+//= require best_in_place
 //= require jquery_ujs
 //= require jquery-ui/sortable
 //= require jquery-ui/effect-highlight
@@ -41,10 +42,14 @@
 //= require selectize
 //= require bootstrap-switch
 //= require bootstrap-datepicker
-//= require rest_in_place
 
 // Jtruncate plugin, http://www.jeremymartin.name/projects.php?project=jTruncate
 // modified by Scott Miller- remove animation, newline for more link
+
+$(document).ready(function() {
+  /* Activating Best In Place */
+  jQuery(".best_in_place").best_in_place();
+});
 
 (function($){
   $.fn.jTruncate = function(opts) {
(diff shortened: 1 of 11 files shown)


app/assets/javascripts/application.js
```diff
@@ -45,6 +46,11 @@
 // Jtruncate plugin, http://www.jeremymartin.name/projects.php?project=jTruncate
 // modified by Scott Miller- remove animation, newline for more link
 
+$(document).ready(function() {
+  /* Activating Best In Place */
+  jQuery(".best_in_place").best_in_place();
+});
+
```
scott: You will need to move this into `app.js` inside the `Helpy.ready()` function so that it works across ajax page change events.  You'll notice that currently it does not work correctly if you click around several times to different tickets and views in the admin.

hin101:
I was gonna style it so the button is a bit highlighted from the page, just a black outline outside the button. What are your thoughts?

scott:
Yeah that sounds good, or maybe even gray just to reduce its prominence a little.  Lets see how it looks.

app/views/admin/topics/_ticket.html.erb
```diff
@@ -6,7 +6,7 @@
 
   <div id="ticket-controls" class="col-md-8 col-xs-12 col-lg-8">
     <h2>
-      <%= render 'admin/topics/ticket_nav_dropdown' %> #<%= @topic.id %>- <%= @topic.name -%>
+      <%= render 'admin/topics/ticket_nav_dropdown' %> #<%= @topic.id %>- <%= best_in_place @topic, :name, as: :input, activator: :span -%>&nbsp;<span class="btn" id="btn_edit_topic_name">Edit</span>
```
CGA1123: Currently this code is passed through the `TopicsController#update` which doesn't check whether the user completing the request is an admin. (This means that any user on the system could change the name of any topic.)

It would be best if the request goes through `Admin::TopicsController#update`, this can easily be done by moving your changes to `TopicsController` over to `Admin::TopicsController` and then changing this line to include `url: admin_topic_path(@topic)`

scott: Good catch @CGA1123 I didn't notice this when I looked at it the first time.  This should be exposed to admins only.

CGA1123:
@hin101 Can you fix the error that Travis is complaining about? 

```app/assets/stylesheets/admin.scss:92 [W] SelectorFormat: Selector `btn_edit_topic_name` should be written in lowercase with hyphens```

File: app/views/admin/topics/_ticket.html.erb [hunk 3]
```diff
@@ -6,7 +6,7 @@
 
   <div id="ticket-controls" class="col-md-8 col-xs-12 col-lg-8">
     <h2>
-      <%= render 'admin/topics/ticket_nav_dropdown' %> #<%= @topic.id %>- <%= @topic.name -%>
+      <%= render 'admin/topics/ticket_nav_dropdown' %> #<%= @topic.id %>- <%= best_in_place @topic, :name, as: :input, url: admin_topic_path(@topic), activator: :span -%>&nbsp;<span class="btn" id="btn-edit-topic-name">Edit</span>
```
scott: Can you make this `class="btn btn-default"` it should match the other buttons and you won't need to style it to 10px.

db/schema.rb
```diff
@@ -16,6 +16,23 @@
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
 
+  create_table "activities", force: :cascade do |t|
```
hin101: This has been fixed in the latest commit as well as the ui changes.

scott:
Awesome!  One last request- Can you figure a way to make the text field fill the container horizontally- maybe use jquery to measure the width of the save and cancel buttons and then subtract it from the total container width. Like this:

screen shot 2016-11-16 at 2 17 59 pm

hin101:
fixed in latest commit

scott:
I was thinking about this, and it seems like maybe using an icon for save and cancel, combined with percentages might be a better way to go.  I will merge this and see what i can do.  Thanks for your work on this @hin101

scott:
Just a retrospective, with BIP editor it was for some reason activating on every link in the same parent container, so any click on a menu would cause the editor to show.  I resolved it by removing the `activate: span`

hin101:
Thanks @scott

PR #432, opened 2016-11-06
Add option to create to 'split' a ticket

### LINKED ISSUE #411: Split Ticket
This will add the ability to split a ticket/topic into two separate topics, by moving posts from one topic to a new one.

It should add an item to "Split" to the post context menu, and upon splitting, create a new topic made up of a copy of the post the split was initiated from.  The name of the new topic should be 'Split from [old topic name]'.  The original topic should still include the post that was split to a new discussion. 

screen shot 2016-11-05 at 10 17 05 am

Upon splitting the post, an internal note should be added to the original topic stating that "A new discussion was created from this one".

CGA1123:
Adds the option to 'split' a ticket into a new discussion.

Will close #411

### DIFF
--- a/app/controllers/admin/topics_controller.rb
+++ b/app/controllers/admin/topics_controller.rb
@@ -316,14 +316,16 @@ class Admin::TopicsController < Admin::BaseController
     parent_topic = Topic.find(params[:topic_id])
     parent_post = Post.find(params[:post_id])
 
-    topic = Topic.new(
+    @topic = Topic.new(
       name: t('new_discussion_topic_title', original_topic: parent_topic.name, default: "Split from #{parent_topic.name}"),
       user: parent_topic.user,
       forum_id: 1,
     )
 
-    if topic.save
-      topic.posts.create(
+    @posts = @topic.posts
+
+    if @topic.save
+      @posts.create(
         body: parent_post.body,
         user: parent_post.user,
         kind: 'first',
@@ -338,7 +340,13 @@ class Admin::TopicsController < Admin::BaseController
       )
     end
 
-    redirect_to admin_topic_path(topic)
+    fetch_counts
+    get_all_teams
+
+    respond_to do |format|
+      format.html { redirect_to admin_topic_path(@topic) }
+      format.js { render 'update_ticket', id: @topic.id }
+    end
   end
 
   private
--- a/app/controllers/admin/topics_controller.rb
+++ b/app/controllers/admin/topics_controller.rb
@@ -318,7 +318,7 @@ class Admin::TopicsController < Admin::BaseController
 
     @topic = Topic.new(
       name: t('new_discussion_topic_title', original_name: parent_topic.name, original_id: parent_topic.id, default: "Split from #{parent_topic.id}-#{parent_topic.name}"),
-      user: parent_topic.user,
+      user: parent_post.user,
       forum_id: parent_topic.forum_id,
       private: parent_topic.private,
     )
--- a/app/controllers/admin/topics_controller.rb
+++ b/app/controllers/admin/topics_controller.rb
@@ -317,7 +317,7 @@ class Admin::TopicsController < Admin::BaseController
     parent_post = Post.find(params[:post_id])
 
     @topic = Topic.new(
-      name: t('new_discussion_topic_title', original_topic: parent_topic.name, default: "Split from #{parent_topic.name}"),
+      name: t('new_discussion_topic_title', original_name: parent_topic.name, original_id: parent_topic.id, default: "Split from #{parent_topic.id}-#{parent_topic.name}"),
       user: parent_topic.user,
       forum_id: parent_topic.forum_id,
       private: parent_topic.private,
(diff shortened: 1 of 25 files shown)


app/controllers/admin/topics_controller.rb
```diff
@@ -312,6 +312,34 @@ def assign_team
     end
   end
 
+  def split_topic
+    parent_topic = Topic.find(params[:topic_id])
+    parent_post = Post.find(params[:post_id])
+
+    topic = parent_topic.forum.topics.new(
+      name: "Split from #{parent_topic.name}",
```
scott: One suggestion would be to use an `i18n` key here so this message can be customized for each locale.

app/controllers/admin/topics_controller.rb
```diff
@@ -312,6 +312,34 @@ def assign_team
...
+      topic.posts.create(
+        body: parent_post.body,
+        user: parent_post.user,
+        kind: 'first',
+        screenshots: parent_post.screenshots,
+        attachments: parent_post.attachments,
+      )
+
+      parent_topic.posts.create(
+        body: 'A new discussion was created from this one',
```
scott: And add a key here instead of hardcoding it

app/controllers/admin/topics_controller.rb
```diff
@@ -312,6 +312,34 @@ def assign_team
...
   end
 
+  def split_topic
+    parent_topic = Topic.find(params[:topic_id])
+    parent_post = Post.find(params[:post_id])
+
+    topic = parent_topic.forum.topics.new(
+      name: "Split from #{parent_topic.name}",
+      user: parent_topic.user
+    )
```
scott: Lets set the `forum_id` to 1 here (which makes the topic private.)  This feature will be used exclusively for splitting private tickets so I don't think it makes sense.

app/views/admin/topics/_post.html.erb
```diff
@@ -10,7 +10,8 @@
 			<ul class="dropdown-menu dropdown-menu-left" role="menu">
 				<li><%= link_to t('edit', default: "Edit"), edit_admin_post_path(post), remote: true %></li>
 				<li><%= link_to t('convert_content', default: 'Turn post into content'), new_admin_doc_path(post_id: post.id, lang: I18n.locale) %></li>
-				<%= content_tag :li, link_to(t('Show Raw Email', default: 'Raw Email'), admin_post_raw_path(post), target: 'blank') if post.raw_email.present? %>
+        <li><%= link_to t('new_discussion', default: 'Create new discussion from this'), admin_split_topic_path(topic_id: post.topic_id, post_id: post.id), method: :post %></li>
```
scott: Can you make this `remote` and handle a js response in the controller?

app/controllers/admin/topics_controller.rb
```diff
@@ -312,6 +312,34 @@ def assign_team
...
+      )
+
+      parent_topic.posts.create(
+        body: 'A new discussion was created from this one',
+        user: current_user,
+        kind: 'note'
+      )
+    end
+
+    redirect_to admin_topic_path(topic)
```
scott: can you have this respond to javascript as well?

CGA1123:
Implemented most of the changes. I'll add the API method later today! 

Thanks for the review!

CGA1123:
@scott That's the API method implemented.

First time using Grape, so let me know if there are any changes necessary!

app/controllers/admin/topics_controller.rb
```diff
@@ -312,6 +312,43 @@ def assign_team
...
 
+  def split_topic
+    parent_topic = Topic.find(params[:topic_id])
+    parent_post = Post.find(params[:post_id])
+
+    @topic = Topic.new(
+      name: t('new_discussion_topic_title', original_topic: parent_topic.name, default: "Split from #{parent_topic.name}"),
+      user: parent_topic.user,
+      forum_id: 1,
+    )
```
scott: I was testing this out and it looks like we also need to explicitly set the `private` flag to `true`.  However, I was also thinking about this and maybe we should just set the `forum_id` and `private` to the `parent_topic` settings.  That way it would work correctly for private tickets, and also for public posts (just in case).

app/controllers/admin/topics_controller.rb
```diff
@@ -312,6 +312,43 @@ def assign_team
...
+      @posts.create(
+        body: parent_post.body,
+        user: parent_post.user,
+        kind: 'first',
+        screenshots: parent_post.screenshots,
+        attachments: parent_post.attachments,
+      )
+
+      parent_topic.posts.create(
+        body: t('new_discussion_post', default: 'A new discussion was created from this one'),
```
scott: I think this would be more usable if we changed it to include the new discussion ID number.  So it would default to "Discussion %{topic_id} was created from this."

app/controllers/admin/topics_controller.rb
```diff
@@ -312,6 +312,43 @@ def assign_team
     end
   end
 
+  def split_topic
+    parent_topic = Topic.find(params[:topic_id])
+    parent_post = Post.find(params[:post_id])
+
+    @topic = Topic.new(
+      name: t('new_discussion_topic_title', original_topic: parent_topic.name, default: "Split from #{parent_topic.name}"),
```
scott: Lets include the ID here- so `"Split from #{parent_topic.id}- #{parent_topic.name}"`

scott:
I thought of one other default for the split `Topic`:  the `topic.user_id` should be set to the `user_id` of the `Post` that was split.  The current functionality is to inherit from the `parent_topic`, but to better support forum discussions I think it is important to inherit from the `Post` instead.

PR #436, opened 2016-11-16
Refactor/topic mailer

### LINKED ISSUE #406: Refactor/rename topic_mailer
This should really be named `post_mailer` since it sends a message to users when an agent replies to them.  Currently Topic.post.last is used when it really should just use a reference to the Post.

scott:
Fixes #406

### DIFF
--- a/app/mailers/post_mailer.rb
+++ b/app/mailers/post_mailer.rb
@@ -1,0 +1,18 @@
+class PostMailer < ActionMailer::Base
+  add_template_helper(ApplicationHelper)
+
+  def new_post(post_id)
+    @post = Post.find(post_id)
+    @topic = @post.topic
+    email_with_name = %("#{@topic.user.name}" <#{@topic.user.email}>)
+    @post.attachments.each do |att|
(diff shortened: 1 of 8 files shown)


CGA1123:
I might have some changes that could be merged in with this PR at [CGA1123:helpy/refactor-mailer](https://github.com/CGA1123/helpy/tree/refactor-mailer)

scott:
Hmm do you know if its possible to diff this in GitHub?

CGA1123:
Yep! [Here](https://github.com/helpyio/helpy/compare/master...CGA1123:refactor-mailer?diff=split&expand=1&name=refactor-mailer)

scott:
My PR is a little different than your refactoring- it actually moves the mailer from the `Topic` to the `Post` model, eliminating a race condition caused by using `Topic.post.last` when it should reference the `Post` by ID.

However, your refactoring looks awesome, submit a PR with it if you haven't yet.

CGA1123:
Have done! #445

PR #450, opened 2016-11-20
Fix deprecation warnings

CGA1123:
Fixes some deprecation warnings.

The `saas-rails` version is bumped as `< 5.0.6` throws:

```
DEPRECATION WARNING: Sprockets method `register_engine` is deprecated.
Please register a mime type using `register_mime_type` then
use `register_compressor` or `register_transformer`.
```

Also #436 led to the following being throw:

```
DEPRECATION WARNING: You are passing an instance of ActiveRecord::Base to `find`.
 Please pass the id of the object by calling `.id`. 
(called from new_post at /Users/christiangregg/dev/helpy/app/mailers/post_mailer.rb:5)
```

### DIFF
--- a/app/models/post.rb
+++ b/app/models/post.rb
@@ -103,8 +103,7 @@ class Post < ActiveRecord::Base
     # Reply from user back to the system
     elsif self.kind == "reply" && self.user_id != self.topic.user_id && self.topic.private?
       I18n.with_locale(self.email_locale) do
-        # NOTE New ticket is misnamed, it should be new-reply
-        PostMailer.new_post(self).deliver_later
+        PostMailer.new_post(id).deliver_later
       end
     end
   end
--- a/Gemfile
+++ b/Gemfile
@@ -6,7 +6,7 @@ gem 'rails', '4.2.7.1'
 gem 'pg'
 gem 'pg_search'
 # Use SCSS for stylesheets
-gem 'sass-rails', '~> 5.0'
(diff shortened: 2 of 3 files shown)


app/models/post.rb
```diff
@@ -103,8 +103,7 @@ def notify
     # Reply from user back to the system
     elsif self.kind == "reply" && self.user_id != self.topic.user_id && self.topic.private?
       I18n.with_locale(self.email_locale) do
-        # NOTE New ticket is misnamed, it should be new-reply
-        PostMailer.new_post(self).deliver_later
+        PostMailer.new_post(id).deliver_later
```
CGA1123: Yeah, passing an instance of `ActiveRecord::Base` was deprecated a while ago (rails/rails@23ffd03ede9b27a6cbc3154aa99c247b718ccdbc).

I noticed travis [throwing these deprecation warnings](https://travis-ci.org/helpyio/helpy/builds/176454552) after #436 was merged.

I could probably actually use `self.id` to keep it consistent with the rest of the file. Although the `self` is redundant.

scott: Thats a good catch because #436 was trying to make the jobs thread safe by passing in IDs to the mailer instead of the object.  I don't think this matters with `suckerpunch` but it does with `sidekiq`.  It still worked correctly passing `self` but passing `id` or `self.id` is much better.

PR #451, opened 2016-11-20
Change Ticket User

### LINKED ISSUE #433: Change Ticket User
Overview:

This enhancement will allow administrators and agents to change the "creator" or `user_id` of a given topic.  Sometimes a need arises to move a topic from one user to another.  This can happen when splitting a topic into multiple discussions, if a user calls in with an issue or an issue is forwarded to the helpdesk.  See  http://support.helpy.io/en/topics/328-change-source-or-asker-of-a-question/posts for further ideas.

Requirements for implementation:

- There should be a button/option to change the creator of a topic.  This should appear only on the first `Post` since that is what determines `Topic` ownership.

screen shot 2016-11-19 at 11 44 29 am

- Upon pressing this button, a ui to search for the user should be presented in a modal window.  This should include a search form and results area so the admin can search for a user by name or email address.
- After clicking the user they would like to transfer the `Topic` to, the `user_id` will be updated for the topic and the first `Post`.
- An internal note should be added to the topic (authored by the agent who changed the Topic creator) stating: "The creator of this topic was changed from [original creator name] to [new creator name]"

CGA1123:
Will close #433

### DIFF
--- a/app/assets/javascripts/admin.js
+++ b/app/assets/javascripts/admin.js
@@ -174,4 +174,8 @@ Helpy.logHistory = function() {
   });
 };
 
+var search_user = function(post_id) {
+  $('#change-user-modal-' + post_id + ' form').submit();
+};
+
 $(document).on('page:change', Helpy.admin);
--- a/app/assets/stylesheets/admin.scss
+++ b/app/assets/stylesheets/admin.scss
@@ -630,3 +630,25 @@ ul.settings-menu {
    height: 0;
  }
 }
+
+.search-results {
+  max-height: 400px;
+  overflow-y: scroll;
+  .change-user-result {
+    &:hover,
+    &:focus {
+      background-color: #ccc;
+      cursor: pointer;
+      cursor: hand;
+    }
+    padding: 5px;
+    min-height: 40px;
+    position: relative;
+    .centering {
+      position: absolute;
+      top: 50%;
+      left: 40px;
+      transform: translateY(-50%);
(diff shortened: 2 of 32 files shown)


CGA1123:
This commit only adds the basic functionality.

Still need to add some more features...

Currently this feature is available for all posts, and I think it might make sense to keep it like that. And to just add a check for if the post is a `first` to update the topic owner as well. 
What do you think about that @scott ?

CGA1123:
search

CGA1123:
Little Gif showing the functionality!

search

scott:
I haven't done a thorough looking through the code yet, but this works really well.  A couple initial thoughts: 

1. There is no internal note added to the thread currently.  We need this-

> An internal note should be added to the topic (authored by the agent who changed the Topic creator) stating: "The creator of this topic was changed from [original creator name] to [new creator name]"

2. The Topic object also caches the "asker" name.  I saw that the `user_id` is changing, but not the cached value.  Not sure if this cached value is actually used anywhere though, but we should update it to match the `user_id`.

3. Didn't see too many tests yet.

Overall, this is awesome and works great!  Great work :)

CGA1123:
@scott 

Will make sure to implemented those changes.

Regarding the caching of the name, currently there is a `before_create` action on the `Topic` model which does the caching. If its not necessary to have it as `before_create`, it could be changed to `after_save`, which would set the cache when the record is created & updated.

Issue #435, opened 2016-11-16
Notify users when ticket is assigned


hvdvegt:
I'd like my agents to be notified when a private ticket is assigned to them. Currently there are two options related to private tickets; get a notification for all new tickets, or a reply in an assigned ticket. Both don't seem to trigger an email when a ticket is assigned.

A have an operator which I'd like to receive a notification for all new tickets and then assign the ticket to the appropriate agent, which is then notified. Is this possible?

scott:
Currently it is not, but adding the functionality would not be hard.  I will add to the roadmap and see who can do the work.

PR #453, opened 2016-11-25
notifies agent when assigned to a private ticket

x6iae:
An enhancement for #435 
========
notifies agent when assigned to a private ticket

### DIFF
--- a/app/views/notification_mailer/ticket_assigned.html.erb
+++ b/app/views/notification_mailer/ticket_assigned.html.erb
@@ -14,7 +14,7 @@
     </p>
     <ol>
       <% @bulk_topics.each do |topic| %>
-        <li><%= topic.name + '  '%><%= link_to 'View', topic %></li>
+        <li><%= topic.name + '  '%><%= link_to t('view_on_site', default: "View on Site"), topic %></li>
       <% end %>
     </ol>
     <p>
--- a/app/views/notification_mailer/ticket_assigned.html.erb
+++ b/app/views/notification_mailer/ticket_assigned.html.erb
@@ -1,0 +1,19 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta content='text/html; charset=UTF-8' http-equiv='Content-Type' />
+  </head>
+  <body <%= 'dir=rtl' if %w(ar dv he iw fa nqo ps sd ug ur yi).include?(@topic.locale) %>>
+    <p>
+      <%= t('above_this_line', default: "Make sure your reply appears above this line") %><br/>
+      =================================================================<br/>
+      <%= t('message_id', default: 'Message ID') %>:<%= @topic.id %></br>
+    </p>
+    <p>
+      <%= t('notify_ticket_assigned', name: @topic.name, default: 'You have been newly assigned to a ticket( %{name} ):') %>
+    </p>
+    <p>
+      <%= t('to_reply', default: 'Reply directly to this email or view this online:') %> <%= admin_topic_url(@topic, host: AppSettings['settings.site_url']) %>
+    </p>
+  </body>
+</html>
--- a/app/views/notification_mailer/ticket_assigned.html.erb
+++ b/app/views/notification_mailer/ticket_assigned.html.erb
@@ -10,7 +10,7 @@
       <%= t('message_id', default: 'Message ID') %>:<%= @topic.id %></br>
     </p>
     <p>
-      <%= t('notify_ticket_assigned', name: @topic.name, default: 'You have been newly assigned to a ticket( %{name} ):') %>
+      <%= t('notify_ticket_assigned', bulk: @bulk ? ', and some others' : '', name: @topic.name, default: 'You have been newly assigned to a ticket( %{name} ) %{bulk}:') %>
     </p>
     <p>
       <%= t('to_reply', default: 'Reply directly to this email or view this online:') %> <%= admin_topic_url(@topic, host: AppSettings['settings.site_url']) %>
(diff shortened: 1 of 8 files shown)


CGA1123:
This feature is nicely done!

However, `Admin::TopicsController#assign_agent` uses `Topic#bulk_assign`, which in turn uses [`update_all`](http://apidock.com/rails/v4.2.7/ActiveRecord/Relation/update_all), which **doesn't trigger callbacks**, meaning that the `after_save` callback you are using to detect the change in an assigned user is never fired.

A fix might be to change `Topic#bulk_assign` to iterate over the given topics, and call `update` on each individually. 

Although I'm not sure what @scott thinks about this.

scott:
Yeah the reason we went to `update_all` was to speed up the operation.  See notes in https://github.com/helpyio/helpy/pull/238  I think the solution might be that bulk assignments don't trigger a notification, or we may have to come up with some other way to do it?

CGA1123:
We could just change the notification to be triggered within the `TopicsController#assign_agent`.

Essentially we would pass `@topics`, and `assigned_user` to `NotificationMailer#ticket_assigned ` able to handle multiple topics.

SundayAdefila:
Yeah @CGA1123 , I think I flow with that line of thought as well, because basically, when there is a bulk assign like this, the assignee is the same person, and I think it will be cool to send the assignee a single notification for all the topics as opposed to individual notification for each

So, I think modifying `NotificationMailer#ticket_assigned` to handle both single and multiple `topics`, and then calling it from the `TopicsController#assign_agent` as well as leaving it in the callback for single updates can work well.

SundayAdefila:
@CGA1123 @scott I made updates to this PR to handle the bulk_assignment... is there some other enhancements that you'll like to suggest in addition to this?

CGA1123:
Seems to look good on first glance.

Although, I feel like it may be better to add proper detail about which topics a user has been assigned to, rather than saying "and some others".

You could loop over `@bulk` to give the topic name and a link.

This would also remove the need to check whether more than 1 topic has been assigned, instead you could just put `You have recently been assigned to the following topic(s):` and then loop over `@bulk` as described above.

SundayAdefila:
Great, thanks @CGA1123 
I implemented that now... looping through `bulk_messages` and listing their names in a ordered list, with a link to `View` each topic assigned.

app/views/notification_mailer/ticket_assigned.text.erb
```diff
@@ -0,0 +1,11 @@
+<%= t('above_this_line', default: "Make sure your reply appears above this line") %>
+=================================================================
+<%= t('message_ids', default: 'Message ID(s)') %>:<%= @bulk_topics.collect(&:id).join(', ') %>
+
+<%= t('notify_ticket_assigned', default: 'You have recently been assigned to the following topic(s):') %>
+
+<% @bulk_topics.each do |topic| %>
+  <%= topic.name %>
```
CGA1123: Adding a link in here would be helpful as well.

app/views/notification_mailer/ticket_assigned.html.erb
```diff
@@ -0,0 +1,24 @@
...
+      <%= t('above_this_line', default: "Make sure your reply appears above this line") %><br/>
+      =================================================================<br/>
+      <%= t('message_ids', default: 'Message ID(s)') %>:<%= @bulk_topics.collect(&:id).join(', ') %></br>
+    </p>
+    <p>
+      <%= t('notify_ticket_assigned', default: 'You have recently been assigned to the following topic(s):') %>
+    </p>
+    <ol>
+      <% @bulk_topics.each do |topic| %>
+        <li><%= topic.name + '  '%><%= link_to 'View', topic %></li>
```
CGA1123: The `View` string should probably be i18n'd.

CGA1123:
@SundayAdefila Looking good, just a couple of small things that could improve this, which I've outlined in the review.

app/views/notification_mailer/ticket_assigned.text.erb
```diff
@@ -5,7 +5,7 @@
 <%= t('notify_ticket_assigned', default: 'You have recently been assigned to the following topic(s):') %>
 
 <% @bulk_topics.each do |topic| %>
-  <%= topic.name %>
+  <%= topic.name + '  '%><%= link_to t('view_on_site', default: "View on Site"), topic %>
```
CGA1123: Same goes here with `' '`.
Also the `link_to` helper returns the html anchor tag (`<a href=.....>`)
We just want to insert the `admin_topic_url(topic, host: AppSettings['settings.site_url'])`

Also, the `Reply directly to this email or view this online` line should probably only be shown if the assignment is for 1 topic. (line 11 of this file)

SundayAdefila: Oh, dang! 
Yeah, thanks.

Also, I think completely removing the `Reply directly to this email or view this online` may be better since each of the assigned topics already have a `View on Site` link, and this will be the save as the link for the reply even when the assignment is for 1 topic.
Thoughts?

CGA1123: Yeah that seems reasonable.

The only problem I can see is that what happens if someone does try to reply from the email...

Is there any way to ignore a reply if it is a "bulk" notification? Maybe @scott knows a way?

SundayAdefila:
Hey @CGA1123 @scott Happy new year guys... Any new update on this?

config/locales/en.yml
```diff
@@ -226,6 +226,7 @@ en:
   account_generated: "An account has been generated for you. To sign in, please use this link:"
   above_this_line: "Make sure your reply appears above this line"
   message_id: "Message ID"
+  message_ids: "Message ID(s)"
```
muescha: new i18n keys should be copied to all languages files as comment to mark missing translations

x6iae:
Is this relevant at all again? I'm thinking of closing it out!

Issue #398, opened 2016-10-19
Prevent deleting system forums and categories


scott:
Should not be able to delete "common replies" category, or system forums (which are currently hidden anyways.)

rschaerer:
:boom: In addition to this, there are two potential data leaks for posts being made in the common replies category:

1.  If someone sets the common replies category to *active* (and optional also enables frontpage) the entries show up in the Knowledgebase ([Source: helpy channel](https://helpyio.slack.com/archives/general/p1476893339000027))
2. If someone sets a post's **Status: Published** and not **Draft** made in the the common repiles category, this post is being exposed by explicit URL (but it keeps not accessible through navigation from the knowledgebase) ([Source: helpy channel](https://helpyio.slack.com/archives/general/p1476893601000029))

hin101:
I can take a look at this for you

scott:
Awesome!

hin101:
Pull Request #459

rschaerer:
Hi @hin101 thanks for the PR #459 ! From your PR description I read that you didn't tackle the two issues I posted on [19 Oct](https://github.com/helpyio/helpy/issues/398#issuecomment-254865864)? (which are not related to the actual issue #398 in the first place and should have been posted as new issue anyways).

If you confirm this, I'll open a new issue with the [data leak](https://github.com/helpyio/helpy/issues/398#issuecomment-254865864) finding.

hin101:
Hi @rschaerer I have hidden it from the front end also, but I haven't tackled 

If someone sets a post's Status: Published and not Draft made in the the common repiles category, this post is being exposed by explicit URL (but it keeps not accessible through navigation from the knowledgebase) (Source: helpy channel)

rschaerer:
@hin101 : could the switches published & draft also made hidden for new posts and editing of posts in the common replies category? What do you think?

PR #459, opened 2016-11-29
Prevent delete system forums and categories

hin101:
I have hidden common replies from the front end, let me know if this is ok or whether we should handle it from the backend.


scott:
Thats a good start, but I think this logic really should be handled by the model.  Having it in the view code is fine for suppressing the option in the UI, but the actual deletion should be encapsulated in a `before_destroy` callback that checks the name of the forum and rejects the deletion accordingly.

Also, a better way to handle the view is by creating a new scope on the Category model, for `public` or `viewable` or something like that. You will probably want that on the show method as well to make sure people (or more likely bots) don't just type in the ID to access the page directly.

Can you check to make sure that the individual docs in the category don't also load?

app/views/categories/_left_nav.html.erb
```diff
@@ -10,14 +10,16 @@
 <div id="left-col" class="col-md-3 hidden-sm hidden-xs affix affix-top">
   <ul class="list-unstyled border-right">
   <% @categories.with_translations(I18n.locale).each do |category| %>
```
scott: You could define a new scope here like `@categories.public.with_trans...` and avoid the `if..end`

app/models/category.rb
```diff
@@ -47,5 +50,8 @@ def read_translated_attribute(name)
     globalize.stash.contains?(Globalize.locale, name) ? globalize.stash.read(Globalize.locale, name) : translation_for(Globalize.locale).send(name)
   end
 
+  def is_not_common_replies?
```
scott: I think for future extensibilty it might be better to use a more generic name- ie. `non_deleteable?` or `protected?`  That way in the future if there are other categories to be protected, they can be added to this method.

app/views/admin/categories/_cat.html.erb
```diff
@@ -12,7 +12,7 @@
       <span id="row-<%= cat.id %>" data-toggle="dropdown" aria-expanded="false" class='btn dropdown-toggle glyphicon glyphicon-align-justify'></span>
       <ul class="dropdown-menu dropdown-menu-right" role="menu">
         <li><%= link_to t(:edit, default: 'Edit'), edit_admin_category_path(id: cat.id, lang: I18n.locale) %></li>
-        <li><%= link_to t(:delete, default: 'Delete'), admin_category_path(id: cat.id), data: {confirm: t(:delete_confirm, default: 'Please confirm you really want to DELETE this, and all articles')}, method: :delete, remote: true, class: 'less-important' %></li>
+        <% unless cat.name == "Common Replies" %><li><%= link_to t(:delete, default: 'Delete'), admin_category_path(id: cat.id), data: {confirm: t(:delete_confirm, default: 'Please confirm you really want to DELETE this, and all articles')}, method: :delete, remote: true, class: 'less-important' %></li><% end %>
```
scott: I would just use the method above here. ie. `unless cat.protected?`

scott:
I created PR #523 handle the common replies privacy issue.  The original issues in #398 still exist and need to be submitted in a new PR (I think this will close when I merge #523)

PR #461, opened 2016-11-30
Return if email is invalid

scott:


lib/email_processor.rb
```diff
@@ -6,6 +6,13 @@ def initialize(email)
   end
 
   def process
+
+    # Guard clause to prevent ESPs like Sendgrid from posting over and over again
+    # if the email presented is invalid and generates a 500.  Returns a 200
+    # error as discussed on https://sendgrid.com/docs/API_Reference/Webhooks/parse.html
+    # This error happened with invalid email addresses from PureChat
+    return if @email.from[:email].match(/\A[^@\s]+@([^@\s]+\.)+[^@\s]+\z/).blank?
```
lccezinha: There is a way to put all this logic inside a method in `@email` class ? And then here call something like `return unless @email.valid?`

CGA1123: That would involve monkey patching `Griddler::Email`. And I don't think that's an ideal solution.

Another option would just be to extract this to a private method `invalid_email?`

scott: I was gonna reply that it would require monkey patching.  @CGA1123 where do you propose putting that method?  Just in `email_processor.rb`? That would help with readability but not so much with sharing the method.

CGA1123: Yeah, I was thinking just within the `EmailProcessor`.

Is there anywhere else where this code needs to be used?

scott: Well its just a quick and dirty email validator in case the email header uses an invalid email address, which shouldn't happen (but did)

scott: I think with the comment it is clear enough what is going on.  This is such an edge case anyways.

PR #463, opened 2016-12-02
Truncate Post Body

### LINKED ISSUE #425: Truncate long posts instead of blocking them
Currently Helpy imposes a limit of 10,000 characters on `Posts`.  If the message is longer, it fails validation and cannot be created.  This is bad because if it is created via an email with a lot of quoted text, the `Post` will not be created, and fail silently.  Instead the better behavior would be to simply limit the character stored.

johnthepink:
This truncates the post body to 10,000 characters during validation. Fixes #425

### DIFF
--- a/app/models/post.rb
+++ b/app/models/post.rb
@@ -26,6 +26,7 @@ class Post < ActiveRecord::Base
   mount_uploaders :attachments, AttachmentUploader
 
   validates :body, length: { maximum: 10_000 }
+  before_validation :truncate_body
   validates :kind, :user, :user_id, :body, presence: true
 
 
@@ -112,5 +113,13 @@ class Post < ActiveRecord::Base
     return I18n.locale if self.kind == 'first'
     self.topic.locale.nil? ? I18n.locale : self.topic.locale.to_sym
   end
+
+  private
+
+    def truncate_body
+      if not self.body.blank?
+        self.body = self.body.truncate(10_000)
+      end
+    end
 
 end
--- a/app/models/post.rb
+++ b/app/models/post.rb
@@ -117,9 +117,7 @@ class Post < ActiveRecord::Base
   private
 
     def truncate_body
-      if not self.body.blank?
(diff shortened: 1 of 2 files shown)


app/models/post.rb
```diff
@@ -113,4 +114,12 @@ def email_locale
     self.topic.locale.nil? ? I18n.locale : self.topic.locale.to_sym
   end
 
+  private
+
+    def truncate_body
```
johnthepink: Totally. I'll get a fix going this evening. Thanks!

test/models/post_test.rb
```diff
@@ -26,7 +26,6 @@ class PostTest < ActiveSupport::TestCase
   should validate_presence_of(:body)
   should validate_presence_of(:kind)
   should validate_presence_of(:user)
-  should validate_length_of(:body).is_at_most(10_000)
```
CGA1123: Can this validation test not be kept?

app/models/post.rb
lccezinha: Thanks bro!

test/models/post_test.rb
johnthepink: @CGA1123 I intended to leave it in, but it fails now that we have the truncation. That test tries to set the body to 10,0001 "x"s:

```
Expected errors to include "is too long (maximum is 10000 characters)" when body is set to [insert 10,001 "x"]
```

Is there a way to bypass the `truncate_body` method for that test?

johnthepink:
@lccezinha @CGA1123 pushed some updates

test/models/post_test.rb
CGA1123: @johnthepink 

Yeah, removing it is the only option I suppose, the new `Should truncate body length if greater than 10,000` test covers this validation anyways.

CGA1123:
LGTM!

lccezinha:
LGTM!

Issue #462, opened 2016-12-01
Enhancements to change user feature


scott:
After getting to actually use the new "change ticket author(user)" feature, I have discovered a couple things that would make it more useful:

1. Change focus to the search bar once the modal is open. 

2. Offer the option to add a user if nothing matching is found in the search.  
- The idea is that if no results are found, there should be a link to add the user
- Instead of "No Results Found," switch to "No matching user found, create them?"
- Upon clicking the "create them" link, it creates the user, using either the name or email that was used to search
- If it was a name search, the email requirement should be satisfied with a placeholder
- Once the user is saved, they should be associated with the Topic#user_id

3. Make the selection of users keyboard friendly, by letting the user arrow up and down and hit enter to select the correct username.  That is how the KB search works on the main page.  At first I thought you used the JqueryUI plugin which makes this pretty easy, but on further review it looks like you built it by hand which might make this harder.  It kinda works now with tabbing though, so this is a really low priority item compared to the others.

PR #464, opened 2016-12-03
Enhance Change User Feature

CGA1123:
PR implementing features outlined in #462

### DIFF
--- a/app/views/admin/posts/_new_user_form.html.erb
+++ b/app/views/admin/posts/_new_user_form.html.erb
@@ -1,8 +1,9 @@
+<div class="add-form">
 <h3>Create new user</h3>
-
-<%= simple_form_for(@user, url: admin_new_user_path, remote: true, validate: true, authenticity_token: true, html: { class: 'form-vertical', method: 'post' }) do |f| -%>
+<%= simple_form_for(@user, url: admin_new_user_path, remote: true, validate: true, authenticity_token: true, html: { class: 'form-vertical', method: 'post', id: 'new_user' }) do |f| -%>
   <%= hidden_field_tag 'post_id', post_id %>
   <%= f.input :email, autofocus: true, class: 'disable-empty' %>
   <%= f.input :name, class: 'disable-empty' %>
-  <%= f.submit 'Create & Set owner' %>
+  <%= f.submit 'Create & Set owner', class: 'btn btn-warning disableable', disabled: true %>
 <% end %>
+</div>
--- a/app/views/admin/posts/_new_user_form.html.erb
+++ b/app/views/admin/posts/_new_user_form.html.erb
@@ -1,0 +1,8 @@
+<h3>Create new user</h3>
+
+<%= simple_form_for(@user, url: admin_new_user_path, remote: true, validate: true, authenticity_token: true, html: { class: 'form-vertical', method: 'post' }) do |f| -%>
+  <%= hidden_field_tag 'post_id', post_id %>
+  <%= f.input :email, autofocus: true, class: 'disable-empty' %>
+  <%= f.input :name, class: 'disable-empty' %>
+  <%= f.submit 'Create & Set owner' %>
+<% end %>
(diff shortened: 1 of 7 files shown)


scott:
I tested this and it seems to work well, but can you fix the one failing check so I can merged it cleanly?

CGA1123:
I can clean up the failing checks.

However, I seem to be having a problem with the button remaining disabled even after having inserted valid data?

scott:
Do you want me to take a look at the disabled button thing?

CGA1123:
hey @scott, yeah please. It has something to do with the `client_side_validation` gem I think.

When I wrote that comment I hadn't pushed the changes I made up here, which is why it would have been working for you.

scott:
Is it pushed now?

CGA1123:
Yeah!

app/views/admin/posts/_new_user_form.html.erb
```diff
@@ -0,0 +1,9 @@
+<div class="add-form">
+<h3>Create new user</h3>
+<%= simple_form_for(@user, url: admin_new_user_path, remote: true, validate: true, authenticity_token: true, html: { class: 'form-vertical', method: 'post', id: 'new_user' }) do |f| -%>
+  <%= hidden_field_tag 'post_id', post_id %>
+  <%= f.input :email, autofocus: true, class: 'disable-empty' %>
+  <%= f.input :name, class: 'disable-empty' %>
```
scott: Same here- when you look in chrome dev tools you will see that the class is not added to the input.

app/views/admin/posts/_new_user_form.html.erb
```diff
@@ -0,0 +1,9 @@
+<div class="add-form">
+<h3>Create new user</h3>
+<%= simple_form_for(@user, url: admin_new_user_path, remote: true, validate: true, authenticity_token: true, html: { class: 'form-vertical', method: 'post', id: 'new_user' }) do |f| -%>
+  <%= hidden_field_tag 'post_id', post_id %>
+  <%= f.input :email, autofocus: true, class: 'disable-empty' %>
```
scott: This is incorrectly formatted- it should use `input_html: { class: 'disable-empty' }`

config/routes.rb
```diff
@@ -103,6 +103,8 @@
     put 'users/invite_users' => 'users#invite_users', as: :invite_users
 
     post 'search/users' => 'posts#search', as: :user_search
+    get  'search/new_user' => 'posts#new_user', as: :new_user
+    post  'search/new_user' => 'posts#change_owner_new_user'
```
scott: This was a little confusing to me, i expected the url to be `/posts/new_user` since thats where the method is.

app/views/admin/posts/new_user.js.erb
```diff
@@ -0,0 +1,2 @@
+$('#change-user-modal-<%= @post_id %> .search-results').html("<%= j render 'admin/posts/new_user_form', post_id: @post_id %>");
+$('#new_user').enableClientSideValidations();
```
scott: This needs to call the global `Helpy.ready();` function after it appends the form, or else the disabled listener does not get added to the new form.

scott:
There is a bigger issue here that once the above things are fixed- the script is still looking for the presence of text in the `post_body` field.  This means that even with the above changes, the `submit` still will be disabled unless you put something in the reply form.  This is not ideal, and is a side effect of the disabled thing really being intended for preventing blank posts.  It predates the addition of client side validations- ie once we added CSV it didn't really need the disabled submit buttons any more.

I think a better solution is to just strip out the `disableable` functionality completely from this form and just let client side validations handle it.  Sure the button will not appear disabled, but the form won't submit with invalid data.

CGA1123:
@scott okay, will push the changes soon.

Thanks for the review!

scott:
Would you say this is good to merge?

CGA1123:
Yeah imo.

Issue #442, opened 2016-11-18
Reject attachment files that are of an unrecognized format


scott:
See #427

Helpy should reject `Post` attachments that have no extension, or that are of an unsupported file type.  A meaningful error message should be displayed to the user.

luisalduucin:
@scott I want to work on it

scott:
@luisalduucin are you working on this?

PR #468, opened 2016-12-06
Reject unsupported file types in attachments

iskvmk:
Fixes https://github.com/helpyio/helpy/issues/442
Error message is displayed if user upload file with unsupported format.

### DIFF
--- a/app/controllers/topics_controller.rb
+++ b/app/controllers/topics_controller.rb
@@ -141,8 +141,8 @@ class TopicsController < ApplicationController
       @topic.user_id = @user.id
     end
 
-    if @user.save && @topic.save
-      @post = @topic.posts.create(
+    if @user.save && @topic.valid?
+      @post = @topic.posts.build(
         :body => params[:topic][:posts_attributes]["0"][:body],
         :user_id => @user.id,
         :kind => 'first',
@@ -150,7 +150,7 @@ class TopicsController < ApplicationController
         :attachments => params[:topic][:posts_attributes]["0"][:attachments])
     end
 
-    if @topic.save
+    if @topic.save && @post.save
       if built_user == true && !user_signed_in?
         UserMailer.new_user(@user.id, @token).deliver_later
       end
@@ -169,8 +169,7 @@ class TopicsController < ApplicationController
         @widget = true
         render 'new', layout: 'widget'
       else
-        flash[:error] = @topic.errors.full_messages.first
-        redirect_to new_topic_path
+        render :new
       end
     end
 
--- a/app/controllers/topics_controller.rb
+++ b/app/controllers/topics_controller.rb
@@ -148,7 +148,11 @@ class TopicsController < ApplicationController
         :kind => 'first',
         :screenshots => params[:topic][:screenshots],
         :attachments => params[:topic][:posts_attributes]["0"][:attachments])
+    end
 
+    @topic.errors.add(:attachments, "Not valid format") unless @post.errors.empty?
+
+    if @topic.save
       if built_user == true && !user_signed_in?
         UserMailer.new_user(@user.id, @token).deliver_later
       end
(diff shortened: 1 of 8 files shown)


app/controllers/topics_controller.rb
```diff
@@ -148,7 +148,11 @@ def create
         :kind => 'first',
         :screenshots => params[:topic][:screenshots],
         :attachments => params[:topic][:posts_attributes]["0"][:attachments])
+    end
+
+    @topic.errors.add(:attachments, "Not valid format") unless @post.errors.empty?
```
CGA1123: Is this handling of the error not better placed in the model?

iskvmk: You're right, moved it to the model.

CGA1123:
I still think that the whole validation process should be moved to the model. 

There are multiple controllers where new posts, with attachment/screenshots can be uploaded.
Keeping the whole validation in the model, through a custom validator, or otherwise, would be a neat way to implement the check.

It would also mean that no controllers would really need to be changed.

If a user tries to upload an unsupported file type, the `Post#save` would fail, and the errors would be added by the model, and then displayed by the view. Essentially the controller shouldn't need to explicitly call validation.

The current `validate_attachment_format`  is a good start, however, it adds an attachment error message if **any** other error has been raised, which I don't think is the intended behaviour.

Also, its probably better to display validation errors in the form, instead of through the flash alert.
Doing it that way means that you can show all the errors, rather than just one.

app/controllers/topics_controller.rb
```diff
@@ -150,7 +150,7 @@ def create
         :attachments => params[:topic][:posts_attributes]["0"][:attachments])
     end
 
-    @topic.errors.add(:attachments, "Not valid format") unless @post.errors.empty?
+    @topic.validate_attachment_format(@post.errors)
```
scott: This is better, but is still calling it inline in the controller, rather than in the validates callback of the model itself.  You can basically take this same method and apply it as a custom validator: http://guides.rubyonrails.org/active_record_validations.html#performing-custom-validations

CGA1123:
Hey @iskvmk, you had any chance to look at moving some of this over to the `Topic` model?

iskvmk:
Hi!
Moved validation to the model. 
Will try to finish with error displaying in the form today, faced some problems there.

iskvmk:
All in all there was no need to add errors to topic model and validate it since topics post contains validation error.

CGA1123:
Looks good, any chance you could write some tests to cover these changes?

iskvmk:
Added tests to cover changes, @CGA1123 please review.

CGA1123:
Hey @iskvmk, feature is looking good.

A few things I noticed:

- When uploading a file that does not have an extension, the app `500` errors, ideally this would be dealt with gracefully, in the same manner that a file with a non whitelisted extension would be.
- Uploading a file with a non whitelisted extension reloads the page, but does not alert the user of this failure, or why it happened.
- Also, a failure to save results in a full `HTTP` request, if there is a way you could also add the possibility of returning a `JS` response on failure, it would speed up the whole cycle.

iskvmk:
Hey @CGA1123 
This feature works only for users with `user` role. It could be done in the same manner for admins, but unfortunately I don't have time for this at the moment.
Rails doesn't support uploading files with `remote: true` option, so there is no built-in functionality to return JS response on failure. But it could be done by `remotipart` gem. Is it worth to add new gem and rewrite forms to show JS response?

scott:
I actually tried to get the remotipart gem working but for some reason I was unable to get it working so I just send the file through a standard request.  It would be a great improvement if you or someone else were able to sort this out.

config/locales/en.yml
```diff
@@ -370,3 +370,6 @@ en:
       save: "Save %{model}"
       new: "New %{model}"
       delete: "Delete %{model}"
+  errors:
+    messages:
+      extension_whitelist_error: 'format is not valid'
```
muescha: i18n messages should be copied to all languages files as comment to mark missing keys

PR #471, opened 2016-12-10
Add params filter for team/group

scott:
Add ability to filter the discussion list by team

### DIFF
--- a/app/controllers/admin/topics_controller.rb
+++ b/app/controllers/admin/topics_controller.rb
@@ -40,11 +40,7 @@ class Admin::TopicsController < Admin::BaseController
     if current_user.is_restricted? && teams?
       topics_raw = Topic.all.tagged_with(current_user.team_list, any: true)
     else
-      if params[:team].present?
-        topics_raw = Topic.all.tagged_with(params[:team], any: true)
-      else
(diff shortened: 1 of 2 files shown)


app/controllers/admin/topics_controller.rb
```diff
@@ -40,7 +40,11 @@ def index
     if current_user.is_restricted? && teams?
       topics_raw = Topic.all.tagged_with(current_user.team_list, any: true)
     else
-      topics_raw = Topic
+      if params[:team].present?
```
lccezinha: Hey @scott what do you think about a simple syntax improve here ? Something like: 

```ruby
if current_user.is_restricted? && teams?
  topics_raw = Topic.all.tagged_with(current_user.team_list, any: true)
else
  topics_raw = params[:team].present? ? Topic.all.tagged_with(params[:team], any: true) : Topic
end
```

Is there some test to cover this fix ? What do you think to add ?

scott: Yes thats a good solution.  That scope should have test coverage from the `acts_as_taggable_on` gem, but I should probably add one for the controller when params are present, so good idea there!

lccezinha:
Nice!

Issue #447, opened 2016-11-18
i18n strings for flagging


scott:
The new flagged posts functionality still needs to be converted to use i18n token and strings.

PR #475, opened 2016-12-15
I18n for flag

hin101:
Added i18n translations for flag feature.


hin101:
Reference to issue #447

app/views/posts/_post.html.erb
```diff
@@ -71,14 +71,14 @@
     <div class="modal-content">
       <div class="modal-header">
         <button type="button" class="close" data-dismiss="modal" aria-label="Close"><span aria-hidden="true">&times;</span></button>
-        <h4 class="modal-title">Flag Post for Review</h4>
+        <h4 class="modal-title"><%= t(:flag_post_for_review) %></h4>
```
scott: We generally provide a `default` for each of the `i18n` strings that is equal to what is provided in the `en.yml` file.

app/views/posts/_post.html.erb
```diff
@@ -71,14 +71,14 @@
...
       </div>
 
       <div class="modal-body row">
         <div class="col-md-12">
         <%= bootstrap_form_for :flag, :url => post_flags_path(post) do |f| -%>
-          <%= f.text_area :reason, rows: 6, cols: 160, placeholder: "Describe the reason for flagging this post", skip_label: true -%>
-          <%= f.submit 'Flag', class: 'btn btn-warning' -%>
+          <%= f.text_area :reason, rows: 6, cols: 160, placeholder: t(:flag_post_placeholder), skip_label: true -%>
+          <%= f.submit t(:flag), class: 'btn btn-warning' -%>
```
scott: Same here, can you add `default:`

config/locales/en.yml
```diff
@@ -82,6 +82,12 @@ en:
   get_your_copy: "Grab your copy on GitHub"
   version: "Version:"
 
+  # flag
+  flag_for_review: "Flag for Review"
+  flag_post_for_review: "Flag Post for Review"
+  flag_post_placeholder: "Describe the reason for flagging this post"
+  flag: "Flag"
+
```
scott: Nice work! Can you also add these keys to the other `yml` files and just comment them out to indicate that they need to be translated?

Issue #441, opened 2016-11-18
Merge Tickets


scott:
This enhancement will make it possible to merge two or more topics into a single new discussion.  From the topics overview page, the agent will be able to select 2 or more topics and then click on 'Merge' to combine the messages.

Requirements and workflow:

- From a Topics overview, the agent should be able to choose two or more topics to merge (existing functionality).
- An option to "Merge" will be presented in the lower menu area (see screenshot).

![screen shot 2016-12-01 at 12 25 00 pm](https://cloud.githubusercontent.com/assets/14369/20808825/4da907e6-b7c1-11e6-9cce-ba4f265de049.png)

- After clicking merge, a dialogue should be presented, confirming the desire to merge the Topics.
- Upon confirming, a new Topic should be created, by creating a copy of the chronologically **first** Topic selected.
- The title of the new Topic should be retained, but prepended with "MERGED: "
- The Posts from each of the selected topics will be copied into the new Topic.
- The selected Topics will be moved to trash.
- An internal note should be written to the new merged Topic stating "%{topic.id}-%{topic.name}, etc. were merged to create this discussion"

hin101:
Ill pick this one up :)

PR #483, opened 2016-12-25
[WIP]Merge tickets

hin101:
Functionality done styling left


hin101:
Issue #441

CGA1123:
Adding i18n strings to this would be ideal!

app/controllers/admin/topics_controller.rb
```diff
@@ -366,6 +366,49 @@ def split_topic
...
+          @posts.create(
+          body: p.body,
+          kind: p.kind,
+          user: p.user,
+          screenshots: p.screenshots,
+          attachments: p.attachments,
+          )
+        end
+
+        topics_merged << "%#{t.id}-%#{t.name}\n"
```
scott: You don't actually need the % signs, that was actually part of the token in the example I gave.

app/views/admin/topics/_tickets.html.erb
```diff
@@ -68,6 +68,7 @@
           <% end %>
 
         </span>
+        <span class="merge-span" style="display: none;"><%= link_to t(:merge, default: "Merge"), admin_merge_tickets_path(merge:''), remote: true, class: 'multiple-update merge', data: { confirm: "Are you sure you want to merge these tickets?" } %></span>
```
scott: We will need i18n for the confirm message too

app/assets/stylesheets/admin.scss
```diff
@@ -650,3 +650,9 @@ ul.settings-menu {
     }
   }
 }
+
+.merge {
+  color: red;
+  font-weight: bold;
+  text-transform: uppercase;
+}
```
scott: We don't actually need the red, I should have noted in the requirements that it was red just to call attention to the new link.  So in other words, the merge link should look like the others in that menu bar.

app/controllers/admin/topics_controller.rb
```diff
@@ -366,6 +366,49 @@ def split_topic
...
+        logger.info("Id #{t.id}")
+        t.posts.each do |p|
+          @posts.create(
+          body: p.body,
+          kind: p.kind,
+          user: p.user,
+          screenshots: p.screenshots,
+          attachments: p.attachments,
+          )
+        end
```
scott: One issue that I didn't think about in the requirements is that this generates two `Posts` with a `kind: 'first'` which causes problems with the formatting.  In this case, we should change the kind to `reply` for all `Posts` except that first one belonging to  `Topic.first`.  Hope that makes sense.

scott:
When I look at the logs for this I am seeing a weird request- `Started GET "/admin/topics/merge_tickets?locale=en&merge=&topic_ids[]=65&topic_ids[]=98&topic_ids[]=65&topic_ids[]=98"`

Should the `topics_ids[]` appear twice?  Also the `merge` param appears to have no value?

app/controllers/admin/topics_controller.rb
CGA1123: Maybe we should add a validation to check for a single post of kind `first` scoped to topic?

CGA1123:
Should this feature really be using a `GET` request? Would `PUT` or `PATCH` not be more appropriate?

scott:
@CGA1123 The thought crossed my mind as well, I think it should be a `PUT` since it is creating a new ticket from the other two.

app/assets/javascripts/app.js
```diff
@@ -224,17 +224,24 @@ Helpy.ready = function(){
     var output;
     var messages = $('.topic-checkbox:checked').size();
 
+    $('.merge-span').hide();
+    console.log("Update");
+
     switch(messages) {
       case 1:
         output = Helpy.selected[1];
+        $('.merge-span').hide();
```
hin101: This is so the merge link only shows when two or more tickets are selected.

scott: Oh okay, good thinking :)

hin101:
I put it as a get request as I want to redirect back to the new topic

Issue #440, opened 2016-11-17
Add filtering options for admin topics view


scott:
Add the abilty to filter the topics view by date ranges, user, group assignment, etc.  More definition coming, also should be built after #365

PR #485, opened 2016-12-29
Enhanced admin search

scott:
This improves the admin search and adds a date-range filter, and the ability to search by group name as well.  It also humanizes large numbers in the pipeline.  It also adds a JS loader when the view cannot be rendered immediately.


scott:
#440

PR #491, opened 2017-01-04
Implement the `User#active` behaviour

### LINKED ISSUE #476: Inactiveting users
Is the 'Active' checkbox working? 
I tried to set off but it doesn't seems to work. The checkbox is unchecked and in the database the users column 'active' is set as false but the user is still able to log in.

CGA1123:
Only `active` users should be allowed to sign in.

Overrides the `Devise::SessionsController#create` action, checks if a user is active, signs them back out if they are not.

Closes #476

### DIFF
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -1,3 +1,14 @@
 class SessionsController < Devise::SessionsController
   theme :theme_chosen
+
+  def create
+    super do |resource|
+      unless resource.active
+        sign_out(resource)
+        flash[:notice] = nil
+        redirect_to root_path, alert: 'Your account is inactive.'
+        return
+      end
+    end
+  end
 end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -5,7 +5,10 @@ Rails.application.routes.draw do
   get 'widget/' => 'widget#index', as: :widget
   get 'widget/thanks' => 'widget#thanks', as: :widget_thanks
 
-  devise_for :users, skip: [:password, :registration, :confirmation, :invitations], controllers: { omniauth_callbacks: 'omniauth_callbacks' }
+  devise_for :users, skip: [:password, :registration, :confirmation, :invitations], controllers: {
+    omniauth_callbacks: 'omniauth_callbacks',
+    sessions: 'sessions'
+  }
 
   as :user do
     get "/users/invitation/accept" => "devise/invitations#edit", as: :accept_user_invitation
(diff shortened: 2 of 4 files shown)


scott:
This looks great, could you write a test spec to cover the active thing? looks like we never had one before.

CGA1123:
Yep will do.

CGA1123:
Added some basic tests, ideally I would want to check that `current_user` is nil when a inactive user tries to sign in, but I can't seem to find a way to actually have access to that method within the tests... 

It's not a huge problem though.

scott:
@CGA1123 Sorry this took so long, I finally got to UAT on this and discovered a couple problems that will be easy to fix:

- If a user is inactive, they should not be able to reset their password.  Currently they can reset password and regain access to their account.
- If a user is an admin and has active API keys, we need to scope the api access by user activity.

Can you work on these and add tests?

CGA1123:
No worries, will have a look at those changes over the coming days!

Sent from my iPhone

On 20 Jan 2017, at 21:59, Scott Miller <notifications@github.com<mailto:notifications@github.com>> wrote:


@CGA1123<https://github.com/CGA1123> Sorry this took so long, I finally got to UAT on this and discovered a couple problems that will be easy to fix:

  *   If a user is inactive, they should not be able to reset their password. Currently they can reset password and regain access to their account.
  *   If a user is an admin and has active API keys, we need to scope the api access by user activity.

Can you work on these and add tests?

-

CGA1123:
@scott sorry for the delay. Uni work is piling up!

Had a look at the reset password case. The reason they are being allowed to log in after a reset is most likely because the devise option [`Devise.sign_in_after_reset_password`](https://github.com/plataformatec/devise/blob/master/app/controllers/devise/passwords_controller.rb#L37) is set to true. Setting this to false in `initializers/devise.rb` could be a quick fix, but would require a user to sign in again after resetting their password. 

If not I can just create a custom controller.

scott:
Good catch, I didn't think about that config setting.  Easy fix :)

CGA1123:
The only disadvantage to that would be that it is easily changed and explicitly forced behaviour by the application!

neliojrr:
Hi! I was looking at devise documentation and I found this 2 methods `active_for_authentication?` and `inactive_message`:
http://www.rubydoc.info/github/plataformatec/devise/Devise/Models/Authenticatable

Maybe it's a good way to solve this overriding this methods in User model doing something like this:
https://github.com/plataformatec/devise/wiki/How-To:-Customize-user-account-status-validation-when-logging-in

What do you guys think?

CGA1123:
@neliojrr Looks very promising! Would be a much cleaner implementation, nice find!

neliojrr:
I commit the code.. take a look: https://github.com/helpyio/helpy/pull/549

CGA1123:
Closing this in favour of #549! @scott

Issue #495, opened 2017-01-05
"Discussions" should be renamed "Tickets"


scott:
After talking with a number of users, it seems that the word "discussions" creates quite a bit of confusion- many customers do not like the concept of "discussion" because it means an open community thread as opposed to a private ticket.  This change will alter the system to use the word 'tickets' in most places, and a new theme will suffice to handle the community forum.

Dayjo:
👍  think this is very wise

samtuke:
+1

PR #496, opened 2017-01-05
Update 'en' locale to use tickets in place of 'discussions'

scott:
#495


CGA1123:
Is there no way to access I18n string within integration tests? Might be better to use them if possible, rather than hard-coding in the values.

scott:
Its actually a good idea, I don't see why it wouldn't work

PR #514, opened 2017-01-17
populate @posts when creating a new ticket from admin

scott:
This fixes a bug where the new message is not properly displayed in the `show` view.


scott:
Closes #513

Issue #499, opened 2017-01-11
Add support for "VIP customers"


samtuke:
Some customers are very important -- please add the ability to identify them and their tickets so that they can be prioritised.

PR #522, opened 2017-01-20
Adds a priority to the users

scott:
This allows you to specify certain customers as higher (or lower) priority than others.  Values include low, normal, high and vip


scott:
#499

Issue #503, opened 2017-01-12
Question about requests to 3rd parties


disturbio:
Would you accept a patch to make requests to 3rd parties (like google analytics, fonts or jsdelivr) optional? we currently on the process to deploy helpy and our users are very privacy minded, meaning we need to strip all of those. It will be cool if we don't have to maintain those changes only in our own fork.

scott:
Yes this would be a valuable addition.  Send over a PR and we can work a little to integrate it into the overall platform as needed.

PR #532, opened 2017-01-28
only submit to google analytics if id set

azul:
First take on #503

### DIFF
--- a/app/themes/flat/views/layouts/_head.html.erb
+++ b/app/themes/flat/views/layouts/_head.html.erb
@@ -24,13 +24,8 @@
 <%= favicon_link_tag "#{AppSettings['design.favicon']}" %>
 <%= cloudinary_js_config %>
 
-<script data-turbolinks-eval="false">
-(function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
-(i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
-m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
-})(window,document,'script','//www.google-analytics.com/analytics.js','ga');
-ga('create', '<%= AppSettings['settings.google_analytics_id'] %>', 'auto');
-</script>
+<%= render :partial => 'layouts/google_analytics' %>
+
 <script>
 Helpy.domain = '<%= "//" + request.host_with_port %>';
 </script>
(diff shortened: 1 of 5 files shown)


azul:
There's a lot of ways how to approach this. This is what seemed the most simple change to me. I'm happy to modify so it fits your coding style / taste.

app/themes/light/views/layouts/_head.html.erb
```diff
@@ -24,13 +24,16 @@
 <%= favicon_link_tag "#{AppSettings['design.favicon']}" %>
 <%= cloudinary_js_config %>
 
+<% if AppSettings['settings.google_analytics_id'].present? %>
```
lccezinha: This 3 snippets of code to add the GA script are all the same ? If they are, what about move it to a partial and call the partial in all places ? 

This way we can use only one conditional to check and in future if we need to add a new condition to check GA we chance this just in the partial, instead chance in 3 diff parts of code.

What do you think ?

lccezinha:
@azul I gave you and advice, can you check it ?

azul:
@lccezinha thanks! Applied your suggestion.

app/themes/flat/views/layouts/_head.html.erb
```diff
@@ -24,13 +24,8 @@
...
 <%= cloudinary_js_config %>
 
-<script data-turbolinks-eval="false">
-(function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
-(i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
-m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
-})(window,document,'script','//www.google-analytics.com/analytics.js','ga');
-ga('create', '<%= AppSettings['settings.google_analytics_id'] %>', 'auto');
-</script>
+<%= render :partial => 'layouts/google_analytics' %>
```
lccezinha: You can simplify this using just:

```
<%= render 'layouts/google_analytics' %>
```

lccezinha:
@azul I gave you another tip, but this is not so crucial :+1: 

About your changes looks good to me, congratz!

exnihilo-creatio:
Will this not generate an error in [event-tracking.js](https://github.com/helpyio/helpy/blob/master/app/assets/javascripts/event-tracking.js)? The script expects the `ga` function to be defined, which it won't be if there is no Google Analytics ID.

I am looking forward to this merge, I had to remove this manually within my own fork and ran into the event-tracking.js issue.

ThomasSevestre:
Analytics is also used by Staccato.

I proposed these changes in the PR I just closed :
https://github.com/helpyio/helpy/pull/592/files#diff-5ba0b8db693dfd7b9c00e488de832273

PR #536, opened 2017-02-01
Add flash message to admin layout

scott:


app/views/admin/settings/index.html.erb
```diff
@@ -1,27 +1,5 @@
 <% title "#{admin_title} #{t('settings', default: 'System Settings')}" %>
-<!-- <div class="agent-header">
```
lccezinha: Nice clean up!

lccezinha:
LGTM !

Issue #416, opened 2016-10-26
Private / Agents only Knowledge Base


silverskysoft:
Hello,
Is there way add topics in KB so that its accessible only to Agents or logged in users? This is so Agents can learn about internal processes etc.

silverskysoft:
@scott any ideas?

scott:
Currently this is not possible but its a pretty good idea and I will add it to the backlog

Rynaro:
It's a great idea. In many companies that I have a opportunity to contribute, I see, the internal documentation is so important.
I'm working on this feature, because it's useful, and will improve the Helpy experience for agents team.

Rynaro:
We have two options to implement this feature on Helpy.

One of them is creating a _internal_ boolean field on *Category* model, and modify, and create some scopes to virtually separate the *public* categories, and *internal* categories. And following this logic, we will need to adjust the *Doc* model too, to accept new Category flow.

And we have an option to completely separate the model, controllers, and views about *internal knowledgebase*. I think, this option isn't elegant, and creates some similar logic in two places, only for "hide" contents.

I really like the *first* option. 
What you think @scott, and @silverskysoft ?

scott:
I thought about this as well, and thought of the first option too.  I think it makes the most sense.  Then create a simple UI inside of the admin namespace for displaying the the new internal KB.

silverskysoft:
I agree, the first option is good. 
 @Rynaro,  Can you also please consider adding `team_list` to Category Model like how Topic model has? Not sure how much work this may add. But do let us know your thoughts.

scott:
Hey @silverskysoft I have been on vacation so I missed this one, but I am curious of your thoughts/reason for adding `team_list` to the Category model?  Is this your suggested approach for determining which categories are "internal only" and to further offer scoping to individual teams?

silverskysoft:
@scott, yes - as you described it. Given the teams feature exists, its almost natural to think of scoping various elements to teams. The immediate questions were:
1. I have a process A for team A. That process contains restricted info -  for example, login info to internal servers. This usually comes in organizations that have dissimilar teams.
2. General use is to limit access to such Knowledge Base articles to teams. 
3. This could also be used to limit search results to more relevant ones in future. 
I'm still trying out Helpy - so not sure if there's another way to achieve the above scoping.

scott:
[...]

PR #538, opened 2017-02-03
Internal Knowledgebase for agent teams

Rynaro:
Create a new feature, that the agent teams can create internal categories and write your own technical/procedural documentation pages.

I adjusted the few tests that give me a conflict. And the new feature needs your own controllers tests.
I prefer to create different controllers, with some "equal" code, to create granularity to the core. If the Internal Category and/or Internal Docs needs some special features, we can avoid to change the public docs core.

I will work on this feature to improve for Helpy community.

References to #416


scott:
This is a great contribution @Rynaro Thanks for doing it!  I like that you added separate controllers, I think that is the best way to do it too.  Do you plan to write some tests?  I may add some comments as I spend a little more time playing with it, but on first look it seems well done!

Rynaro:
Yes, @scott. I will create some tests for new controllers, and some others for the existent "public controllers", and all controllers that be affected by this feature. 
And I will adjust the code to fix the security offense inside the feature code. 😃

Rynaro:
@scott , I've made a refactor inside Doc, and InternalDoc controllers, to be more concise, and more easy to read.
Using aspect programming concept eliminate a conditional structure, and tests works well too.
Using a helper method to sanitize the Doc content. Breakman now passes test successfully.

app/models/category.rb
```diff
@@ -54,4 +60,12 @@ def non_deleteable?
     return false if name == "Common Replies"
   end
 
+  def is_public_viewable?
```
CGA1123: Would `publicly_viewable?` not be more appropriate?

app/models/category.rb
```diff
@@ -54,4 +60,12 @@ def non_deleteable?
...
+    public && name != "Common Replies"
+  end
+
+  def is_internal_viewable?
```
CGA1123: Same here, `internally_viewable?`

app/models/category.rb
```diff
@@ -33,7 +35,11 @@ class Category < ActiveRecord::Base
   scope :ordered, -> { order('rank ASC') }
   scope :ranked, -> { order('rank ASC') }
   scope :featured, -> { where(front_page: true) }
+  scope :non_featured, -> { where(front_page: true) }
   scope :viewable, -> { where.not(name: 'Common Replies')}
+  scope :is_public, -> { where(public: true) }
+  scope :is_internal, -> { where(public: false) }
```
CGA1123: Feel like maybe these scopes should match with the later methods, as well as the `viewable` scope?
`publicly_viewable` & `internally_viewable`

app/models/category.rb
Rynaro: I agree. `is_` sounds strange for me too. Thanks for the hint! 😄

config/locales/en.yml
```diff
@@ -373,6 +374,9 @@ en:
       root: *root
     result:
       root: *root
+    admin:
+      internal_categories:
+        root: "Internal Categories"
```
muescha: i18n should be added to all language files as comments to mark untranslated keys

config/locales/en.yml
```diff
@@ -333,6 +333,7 @@ en:
         meta_description: Meta Description
         front_page: Front Page
         active: Active
+        public: Public Category Content
```
muescha: i18n should be added to all language files as comments to mark untranslated keys

app/controllers/admin/categories_controller.rb
```diff
@@ -59,13 +59,16 @@ def category_params
     :front_page,
     :active,
     :section,
-    :rank
+    :rank,
+    :team_list,
```
scott: Is this actually used?  I remember we discussed scoping by team, but I didn't see it, or missed it...

app/views/layouts/_admin_header.html.erb
```diff
@@ -25,6 +25,7 @@
           <%= content_tag(:li, link_to(t(:team, default: 'Team'), admin_stats_path(interval: 7))) if tickets? && current_user.is_admin? %>
           <%= helpcenter_menu if (knowledgebase? || forums?) && current_user.is_agent? %>
           <%= content_tag(:li, link_to(t(:content, default: "Content"), admin_categories_path), class:'kblink') if knowledgebase? && current_user.role == 'editor' %>
+          <%= content_tag(:li, link_to(t(:internal_content, default: "Internal Content"), admin_internal_categories_path), class:'kblink') if knowledgebase? %>
```
scott: Not to diminish the importance of this feature, but I think it might be better suited in the "Help" dropdown since I think the primary use case is as an internal wiki or training reference for agents.  Thoughts on this?

config/locales/en.yml
scott: I think just "Public" or "Publicly Available" might be clearer.  Or have a radio button for "Public" or "Internal" instead of a checkbox.

app/controllers/admin/internal_docs_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+    @doc = Doc.find_by(id: params[:id], active: true)
+    redirect_to controller: 'errors', action: 'not_found' if @doc.nil? || !@doc.category.internally_viewable?
+  end
+
+  def generate_page_breadcumbs
+    add_breadcrumb t(:knowledgebase, default: "Knowledgebase"), categories_path
+    add_breadcrumb @doc.category.name, category_path(@doc.category) if @doc.category.name
+    add_breadcrumb @doc.title
+    @page_title = @doc.title
+  end
```
scott: This is a good way to encapsulate this logic!  Although I didn't see the breadcrumbs used in the view?  DId I miss it?  Also, can this be shared in the non-internal `docs_controller`?

app/controllers/docs_controller.rb
```diff
@@ -27,27 +27,49 @@
...
+
+  def define_topic_for_doc
+    if @doc.topic.present?
+      @topic  = @doc.topic
+      @post   = @topic.posts.new
+      @posts  = @topic.posts.ispublic.active.includes(:user) unless @topic.nil?
     else
-      redirect_to controller: 'errors', action: 'not_found'
+      @topic  = Topic.new
+      @post   = Post.new
```
scott: Some great refactoring!

app/views/layouts/_admin_header.html.erb
scott: I also think this should not be conditional not on `knowledgebase?` but perhaps instead based on a settings variable.

muescha: missing i18n for key `:internal_content` (in en.yaml and others)

config/locales/en.yml
Rynaro: @muescha , I really forgot to add this key on others i18n files, I will provide this adjustment on next commit. Thanks! :smiley:

config/locales/en.yml
Rynaro: I think the radio button providing _public_ and _internal_ values, just add more "clean-way" to use this feature. I will provide this change on next commit! Thanks :smiley:

app/controllers/admin/categories_controller.rb
Rynaro: @scott, I just added this featured to create a filter by team, but looking on feature, I think, we need to make a review on this feature on Internal KB.

app/views/layouts/_admin_header.html.erb
Rynaro: Makes sense to move _Internal Content_ to Help dropdown! @scott 

@muescha, I've made a comment on your previous note. Thanks for review my WiP code. :smiley:

config/locales/en.yml
scott: I actually was thinking about this more- do you think we should actually have three states- internal only, public only, and both?  In this case it would work to have two checkboxes (or three radio buttons).  What do you think?

app/controllers/admin/internal_categories_controller.rb
```diff
@@ -0,0 +1,32 @@
...
+    else
+      redirect_to controller: 'errors', action: 'not_found'
+    end
+  end
+
+  private
+
+  def generate_page_breadcumbs
+    add_breadcrumb t(:internal_content, default: "Internal Content"), admin_internal_categories_path
+    add_breadcrumb @docs.category.name, category_path(@docs.first.category) if !@doc.nil? && @doc.first.category.name
```
scott: Still using the external link path here

app/controllers/comments_controller.rb
```diff
@@ -16,7 +16,7 @@ def create
     @post.screenshots = params[:post][:screenshots]
 
     if @post.save
-      redirect_to doc_path(@doc)
+      redirect_to path_based_on_category_public_status(@doc)
```
scott: I think this might be better done by passing a source_of_comment attr instead of looking purely at whether or not it is public.  I was testing it and was surprised to get redirected from the admin experience back out to the public layout after making a comment

app/views/admin/categories/_form.html.erb
```diff
@@ -6,9 +6,14 @@
   <%= f.input :meta_description, label: I18n.translate("activerecord.attributes.category.meta_description", locale: I18n.locale), value: @category.read_translated_attribute(:meta_description) %>
   <% if params[:lang] == "#{I18n.locale}" %>
     <%= f.input :front_page, :label => I18n.translate("activerecord.attributes.category.front_page", locale: I18n.locale), :input_html => {:class => 'no-translate'} %>
+    <%= f.collection_radio_buttons :visibility, [
```
scott: These are really jammed together, can you space them out a bit?

PR #539, opened 2017-02-03
Update Japanese messages.

monaka:


Rynaro:
Has something wrong in line 62 that breaks the YAML schema. Travis can't read your code because of this.
Adjust the key parsing error, and I think Travis build will be successful. :)

monaka:
@Rynaro Thanks for your advice.
The build failure was caused a multi-byte white space. I forgot to check before committing it. 
I fixed there and force-pushed. Please check it again.

monaka:
Added some more traslations.

scott:
Awesome, thanks for providing these updates!

PR #541, opened 2017-02-06
[WIP] allow tickets to be viewed and replied to via secret url

elijh:
This PR adds "codes" to topics, so that tickets may be viewed and replied to via a secret URL.

The desire for this change comes from a need to make "email" an optional field for creating a new ticket when not authenticated (we don't want to require an email address when asking questions of an email provider). 

I don't know if something like this would be welcome in helpy, but if so, then here is one possible solution. This current PR is not particularly clean, but I am pushing it here to stimulate a conversation. 

The approach taken here is to use a random string for the user.email field if email is missing, since there is an index on the email column that requires an unique value.

The disadvantage of this is that a new user is created for every anonymous ticket, and that new user is never destroyed. Before, when email was required, when the same person created a second ticket, the email address could be used to fetch the existing user object to associate with the new ticket.

This is not really a problem, but it does suggest that it would be good to create scripts to clean out old users and old closed private topics.

### DIFF
--- a/app/controllers/posts_controller.rb
+++ b/app/controllers/posts_controller.rb
@@ -41,10 +41,22 @@ class PostsController < ApplicationController
   end
 
   def create
-    @topic = Topic.find(params[:topic_id])
-    @post = @topic.posts.new(post_params)
+    if params[:topic_code]
+      @topic = Topic.find_by_code(params[:topic_code])
+      @post = @topic.posts.new(post_params)
+      if user_signed_in?
+        @post.user_id = current_user.id
+      else
+        @post.user_id = @topic.user.id
+      end
+    elsif params[:topic_id] && user_signed_in?
+      @topic = Topic.find(params[:topic_id])
+      @post = @topic.posts.new(post_params)
+      @post.user_id = current_user.id
+    else
+      raise ArgumentError, 'either topic_code or topic_id must be in params'
+    end
     @post.topic_id = @topic.id
-    @post.user_id = current_user.id
     @post.screenshots = params[:post][:screenshots]
 
     respond_to do |format|
--- a/app/models/auto_user.rb
+++ b/app/models/auto_user.rb
@@ -1,0 +1,45 @@
+#
+# A class for automatically generated users
+#
+
+require 'securerandom'
+
+class AutoUser < User
+  # tell devise that email is optional
+  def email_required?; false; end
+
+  before_validation :signup_guest, :on => :create
+
+  #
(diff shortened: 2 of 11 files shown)


app/models/auto_user.rb
```diff
@@ -0,0 +1,45 @@
...
+    # The only way to remove a validation set in a super class is to modify the
+    # _validators variable. Here we remove the validator set in user.rb on
+    # email. We can leave the ones set by devise, because of email_required?()
+    _validators[:email].delete_if do |v|
+      v.is_a?(ActiveRecord::Validations::PresenceValidator) && v.options.empty?
+    end
+  end
+
+  def signup_guest
+    # the .invalid TLD is reserved for uses like this
```
scott: I like this.  Had not see it defined that way before.

scott:
Hey @elijh thanks for contributing this!  Sorry for the delay getting to review this, I have been completely backed up on another project.

This has been an interesting problem that we have discussed before- #69 and in other places.  I like the secret URL approach because it lets users check their ticket without logging in.  What if we provided a way for users to "claim" their user account by providing a path from the secret url... ie- prompt them to enter their email and then remove the access via secret code?

app/controllers/posts_controller.rb
```diff
@@ -41,10 +41,22 @@ def index
...
+        @post.user_id = current_user.id
+      else
+        @post.user_id = @topic.user.id
+      end
+    elsif params[:topic_id] && user_signed_in?
+      @topic = Topic.find(params[:topic_id])
+      @post = @topic.posts.new(post_params)
+      @post.user_id = current_user.id
+    else
+      raise ArgumentError, 'either topic_code or topic_id must be in params'
```
muescha: missing i18n and more clear message for end user (the links may be destroyed or cut by buggy email link parsers in email clients)

config/locales/en.yml
```diff
@@ -69,6 +69,9 @@ en:
   ticket_number: "Ticket Number"
   assigned_to_label: "Assigned to"
   created_on: "Created on"
+  ticket_secret_url_notice: |
+    You are viewing this ticket using the secret link %{url}.
+    Anyone with this link can view and reply to this ticket.
```
muescha: new keys to en.yml should be copied to all language files as comments to mark untranslated keys

app/controllers/posts_controller.rb
hin101: This error would show on the logs, may be useful to display a flash message here also, so users know there was an error.

Issue #457, opened 2016-11-29
delete a user


hadifarnoud:
sometimes there are newsletters and other spam emails that show up and those are not real users. would be great to be able to delete them

wilsonpeng8:
ive also noticed theres no way to delete a user from the user list

micah:
I was just searching how to delete a user, and I couldn't find one, so I came here to file an issue, and looks like there is one!

tomatoturnip:
Is anyone working on a PR for this? If not, I can start on it.

scott:
I am not aware of anyone working on this.  A PR would be great!

tomatoturnip:
Cool I'll get started on it. Thanks @scott! :)

scott:
Awesome.  I think it would be ideal to put this into a background job once you have it working in principal.

PR #545, opened 2017-02-15
Admin can delete users

tomatoturnip:
This feature allows admins to delete users one at a time. Let me know if you want to batch delete them. Deleting a user also deletes all associated objects (i.e., topics, posts, votes, docs, and api keys). For issue: #457

### DIFF
--- a/app/controllers/api/v1/users.rb
+++ b/app/controllers/api/v1/users.rb
@@ -171,6 +171,17 @@ module API
           present user, with: Entity::User
         end
 
+        # DELETE A USER
+        desc "Delete a user"
+        params do
+          requires :id, type: Integer, desc: "User ID"
+        end
+        delete ":id", root: :users do
+          user = User.find(permitted_params[:id])
+          user.destroy
+          body false
+        end
+
         # INVITE USER
         desc "Invite one or more users to create an account"
         params do
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -90,12 +90,12 @@ class User < ActiveRecord::Base
   paginates_per 15
 
   # Relationships
-  has_and_belongs_to_many :roles
-  has_many :topics
-  has_many :posts
-  has_many :votes
-  has_many :docs
-  has_many :api_keys
+
+  has_many :topics, dependent: :delete_all
+  has_many :posts, dependent: :delete_all
+  has_many :votes, dependent: :delete_all
+  has_many :docs, dependent: :delete_all
+  has_many :api_keys, dependent: :delete_all
   has_attachment  :avatar, accept: [:jpg, :png, :gif]
   is_gravtastic
 
--- a/app/views/admin/users/_user.html.erb
+++ b/app/views/admin/users/_user.html.erb
@@ -23,6 +23,9 @@
       <ul class="dropdown-menu dropdown-menu-right" role="menu">
         <li><%= link_to t(:discussions, default: 'Discussions'), admin_user_path(user), remote: true %></li>
         <li><%= link_to t(:edit_user, default: 'Edit User'), admin_user_path(user, mode: 'edit'), remote: true %></li>
+        <li><%= link_to t(:delete_user, default: 'Delete User'), admin_user_path(user), remote: true,
+                        method: :delete,
+                        data: { confirm: "Are you sure you want to delete #{user.name}?" } if current_user.is_admin? %></li>
       </ul>
     </div>
   </td>
--- a/config/locales/ca.yml
+++ b/config/locales/ca.yml
@@ -160,6 +160,7 @@ ca:
   contact_info: Contacte
   social_info: Social
   edit_user:
+#  delete_user: Delete User
   work_phone: feina
   cell_phone: mòbil
 
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -163,6 +163,7 @@ en:
   contact_info: Contact
   social_info: Social
   edit_user: Edit User
+  delete_user: Delete User
   work_phone: work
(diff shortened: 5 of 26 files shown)


config/locales/en.yml
```diff
@@ -163,6 +163,7 @@ en:
   contact_info: Contact
   social_info: Social
   edit_user: Edit User
+  delete_user: Delete User
```
hin101: we should add this line to all translation files but leave it commented, as this indicates it needs to be translated

hin101:
We should also add this feature in app/controllers/api/v1/users.rb, so that it is available in the API.

config/locales/en.yml
tomatoturnip: Thanks! I'll do that. :)

tomatoturnip:
@hin101 sure no prob, I'll add it to the API

app/controllers/api/v1/users.rb
```diff
@@ -171,6 +171,17 @@ class Users < Grape::API
...
+        desc "Delete a user"
+        params do
+          requires :id, type: Integer, desc: "User ID"
+        end
+        delete ":id", root: :users do
+          user = User.find(permitted_params[:id])
+          user.destroy
+          body false
+        end
+
```
hin101: excellent :)

app/models/user.rb
```diff
@@ -90,12 +90,12 @@ class User < ActiveRecord::Base
...
   # Relationships
-  has_and_belongs_to_many :roles
-  has_many :topics
-  has_many :posts
-  has_many :votes
-  has_many :docs
-  has_many :api_keys
+
+  has_many :topics, dependent: :delete_all
+  has_many :posts, dependent: :delete_all
```
hin101: Do we want to preserve topics and posts, but unassign them? Might be one for Scott to clarify.

app/models/user.rb
```diff
@@ -90,12 +90,12 @@ class User < ActiveRecord::Base
...
+  has_many :votes, dependent: :delete_all
+  has_many :docs, dependent: :delete_all
```
scott: @hin101 @tomatoturnip I actually think docs should be preserved too, since deleting an admin or author could easily lead to inadvertently destroying all docs too.

app/models/user.rb
scott: Yeah I agree.  Its tough to answer for all cases because if you are deleting spam or test users, then you would also want to destroy all dependent topics.  However, for a business, removing history is probably not what is desired.  I think it would be ideal to show a warning screen with a checkbox to allow the topics to be deleted with the user.

app/views/admin/users/_user.html.erb
```diff
@@ -23,6 +23,9 @@
       <ul class="dropdown-menu dropdown-menu-right" role="menu">
         <li><%= link_to t(:discussions, default: 'Discussions'), admin_user_path(user), remote: true %></li>
         <li><%= link_to t(:edit_user, default: 'Edit User'), admin_user_path(user, mode: 'edit'), remote: true %></li>
+        <li><%= link_to t(:delete_user, default: 'Delete User'), admin_user_path(user), remote: true,
+                        method: :delete,
+                        data: { confirm: "Are you sure you want to delete #{user.name}?" } if current_user.is_admin? %></li>
```
scott: As I mentioned above, I think that instead of deleting immediately, we need some kind of intermediate view/modal/warning that says something to the effect of "This user has created X number of tickets.  Do you want to keep these?" 

If the answer is yes, then the user_id should be switched to 2 which is the system user.  An internal note should be added to the history of any topics documenting this change.

app/models/user.rb
muescha: or add a generic dummy user "deleted user' to the topics and posts

app/views/admin/users/_user.html.erb
muescha: missing i18n in this line

config/locales/ca.yml
```diff
@@ -160,6 +160,7 @@ ca:
   contact_info: Contacte
   social_info: Social
   edit_user:
+#  delete_user: Delete User
```
sesharim: Are comments allowed in master branch?

CGA1123: Comments within locale files are added in to mark untranslated strings!

app/views/admin/users/_user.html.erb
tomatoturnip: @scott & @hin101: As a quick fix, an option is to set those dependent active record associations as dependent: :nullify, which would set the foreign keys to nil when a user is destroyed as opposed to assigning them to a dummy/system user. However, that would not give the admin the control to delete those orphaned records completely. I'm working on a commit to give the admin this option but just wanted to bring up the dependent: :nullify option just in case that's preferable.

app/models/user.rb
scott: @muescha I thought about that approach too.

app/views/admin/users/_user.html.erb
scott: Yeah I don't think nullify is a great choice here either.  Looking forward to the commit @tomatoturnip!  Thanks for your contributions!

app/models/user.rb
muescha: or instead of a "dummy user": remove all the user related data and use a "generic name" for that user. otherwise a user can get a big bucket of related posts / topics which are not related.

scott: Thats probably the best solution!

Lubo:
Hello, I see no activity in nearly a month. What's the problem? Can I help with something?

PR #549, opened 2017-02-18
user inactive can not log in

neliojrr:


app/models/user.rb
```diff
@@ -236,6 +236,16 @@ def send_devise_notification(notification, *args)
     devise_mailer.send(notification, self, *args).deliver_later
   end
 
+  # check if user is active or not
+  def active_for_authentication?
+    super and self.active?
```
CGA1123: Travis is complaining about using `and`, `&&` is preffered!

CGA1123:
Seems to work good. Deals with the reset password loophole as well.

app/models/user.rb
```diff
@@ -236,6 +236,16 @@ def send_devise_notification(notification, *args)
...
   end
 
+  # check if user is active or not
+  def active_for_authentication?
+    super && self.active?
+  end
+
+  # message to the user that is not allowed to login
+  def inactive_message
+    "You are not allowed to log in!"
```
muescha: missing i18n

lccezinha: The code is simple enough. It's not necessary to comment it.

scott: I think @muescha is saying to use an i18n string instead of hard coding the message

muescha: but @lccezinha point to an other problem: to reduce comments for just one liners. because the often get out of sync later

scott: Oh yeah thats true @muescha I didn't even pay attention to the comment

scott:
Yeah seems to work well, thanks for helping on this everyone!

muescha:
checked in without i18n :(

scott:
Submit a PR :)  Its an important bug fix that has been waiting for a long time or I wouldn't have.

PR #553, opened 2017-02-20
fix german translations

muescha:


muescha:
@scott when this changes becomes alive at http://demo.helpy.io ?

is there any CI schedule for updating the demo site?

scott:
Hey, there is not a CI for the demo, I update it when I have time.  It should be pretty soon though :)

PR #556, opened 2017-02-21
Fix typo

ellimist:


ellimist:
Anyone knows how this small text fix could break a completely unrelated test?

CGA1123:
@ellimist not sure, having a look at it right now.

CGA1123:
@ellimist 

I think it's because the test relies on the ordering of the response, and the parsing of the JSON response.

In theory the `Hash` data structure is order independent, which would mean that this test can fail trivially. (Not sure tough, need to dig a bit more)

Either way, it isn't your change that caused it.

ellimist:
Yeah, that's what I thought too. Can you try to re-trigger the build job?

ellimist:
Closing and re-opening the pull-request did the trick too 👍  Let's see if the tests pass this time

CGA1123:
Strange, maybe there's something else going on...

ellimist:
@CGA1123 but you were right, the order of the keys is screwing up the tests. The first one failed cause `note` was expected to equal `first`, the second on failed because `note` was expected to equal `reply`.

CGA1123:
@ellimist Ahh well spotted!

CGA1123:
@ellimist Well spotted! Can you open an issue regarding this?

scott:
Yeah this failing test thing has been around for a while.  It only seems to happen once in a while...

scott:
Magically reran the build and its green :)

PR #564, opened 2017-03-04
Remove the redundant setting.

monaka:
`config.action_mailer.raise_delivery_errors` will be `false` by default.

And it will cause some confusion to users
as there have the comment `# config.action_mailer.raise_delivery_errors = false` above it.


scott:
Thanks @monaka this makes sense

PR #576, opened 2017-03-17
Sync some translation keys

Rynaro:
I've made the sync with the latest commit that have added some keys on shortcut modal feature.
The languages English, and Português Brasileiro (Brazilian Portuguese) are fully translated.

Others languages has the keys as comment with the English version of translation values to help the future _translator_.


scott:
This is awesome @Rynaro thanks for doing the syncing too!

PR #577, opened 2017-03-20
Allow Ruby 2.4

excid3:
Aside from some warnings in dependencies, Helpy runs great on Ruby 2.4. This just relaxes the Ruby version dependency to support anything below Ruby 3 as there shouldn't be any changes preventing newer Ruby 2.x support.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ language: ruby
 rvm:
   - 2.2.5
   - 2.3.1
+  - 2.4.0
 
 services:
   - postgresql
--- a/Gemfile
+++ b/Gemfile
@@ -189,4 +189,4 @@ group :production do
   gem 'unicorn'
 end
 
-ruby '>= 2.2', '< 2.4'
+ruby '>= 2.2', '< 3.0'


scott:
Thanks for this @excid3!

scott:
Hey @excid3 since merging in this ruby 2.4 pr, I am getting a failed test assertions in my branches, only for the ruby 2.4 build.  Take for example https://github.com/helpyio/helpy/pull/582  The fail is in the `integration_test_helper` which seems totally unrelated and the method is `sign_in` which is successfully called multiple other times.  Seems like this might be a travis issue, curious if you have any ideas?

excid3:
You know, on my laptop I think I had that same failure when I ran the tests, but on my desktop when I ran them I didn't have that issue.

Interestingly it's failing on `click_on('Sign in')` it looks like. Wonder what the heck is causing that.

scott:
I updated travis to 2.4.1 and it seems to have resolved the issue.  Although I think it may have been a weird circumstance where travis was using the config from the master, but my branches didn't have 2.4 and it was resulting in that strange failed test.  Strange.

excid3:
Yeah, that is odd. Guess we can just keep a cautious eye out for any other things for a while.

PR #578, opened 2017-03-20
Fix Estonian translation

pepzwee:
Added missing translations and fixed some previous translations.


scott:
Thanks for adding these @pepzwee

PR #592, opened 2017-03-29
do not load analytics js if settings.google_analytics_id is blank?

ThomasSevestre:


ThomasSevestre:
There already is a PR for this :
https://github.com/helpyio/helpy/pull/532

Issue #561, opened 2017-03-02
Add ability to invite agents by sharing link, instead of only email


scott:

hin101:
Hi Scott, to invite an agent, does the user have to be an administrator?

scott:
Yes that is correct.

PR #612, opened 2017-04-16
[WIP] Invite agents by sharing link

### LINKED ISSUE #552: Add ability to "resend invite" to previously invited agents
Add a menu item to "resend invite" to previously invited agents/users

hin101:
Closes #552 and #561


scott:
This is good although I noticed the link displayed on the the user page does not work.  The link in the email when you resend it does however.  They are not using the same token for some reason.

app/views/admin/users/_user_info_horizontal.html.erb
```diff
@@ -30,5 +30,14 @@
     LinkedIn: <%= @user.linkedin %><br/>
   </div>
 
+  <% if @user.invited? %>
+    <br />
+    <div id="user-invited" class="col-md-9 col-sm-3 col-xs-6 user-info">
+      <div class="tiny-header"><%= t(:user_invited, default: "User Invited") %>:</div>
+      Invitation URL: <%= accept_invitation_url(@user, :invitation_token => @user.invitation_token, host: AppSettings['settings.site_url']) %>
```
scott: This must either be using an expired token or just the wrong one somehow.

hin101:
Would someone be able to take a look at this, I can't get the token to work.

scott:
I did a little research on this, and it appears to not be possible to do it after the fact.  At the moment you invite a user, there is a `raw_token` available but that is not persisted to the database.

That causes issues with sending emails with `deliver_later` because that attribute is not available on the user model.  @mkumaror1989 built the invite feature and he got around that by storing the link in the text message that is sent.  It reduces security but otherwise you have to send inline which is not good either.

So you can either access the invite text and parse out the URL from there, or create a reset password link and use that: http://www.rubydoc.info/github/plataformatec/devise/Devise/Models/Recoverable

I think using a reset link is preferable because you can set an expiration date, and it does the same thing.

PR #623, opened 2017-05-12
minor: remove unused TopicsController#post_params

### LINKED ISSUE #622: minor: post_params in TopicsController is unused
I don't think this is used anywhere.
https://github.com/helpyio/helpy/blob/master/app/controllers/topics_controller.rb#L158

Confused the heck out of me never the less.

azul:
fixes #622

### DIFF
--- a/app/controllers/topics_controller.rb
+++ b/app/controllers/topics_controller.rb
@@ -155,14 +155,6 @@ class TopicsController < ApplicationController
 
   private
 
-  def post_params
-    params.require(:post).permit(
-      :body,
-      :kind,
-      {attachments: []}
-    )
-  end
-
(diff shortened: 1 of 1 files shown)


azul:
No idea, why the first build fails. Seems unrelated.
I'd prefer the version in the second commit... but i guess that is a matter of taste. I've seen a bunch of require().permit() in the rest of the code though.

scott:
Sometimes we have build failures with Travis.  This happens a lot less frequently with Codeship which I also use.  I restarted that build and expect it will pass.  As for the PR- this looks like some unexpected cruft.  I am still investigating if it might be part of a prior refactoring that didn't get full merged in-  #181 for example.

Issue #632, opened 2017-05-19
Refactor && DRY new user creation method


scott:
As noted in #628 there is a significantly duplicated method for instantiating a new user that is repeated around the codebase.  Would be best to refactor these into one, drying up the code.

scott:
Some work was done on there here #181

GPrimola:
Hi again @scott! I'm going to work on this issue!

scott:
Thats great!

PR #635, opened 2017-05-20
PR for issue 632 - Refactor && DRY new user creation method

GPrimola:
Refactor of email_processor#create_user to use user#register. Now user#register is also using the previous existent user#signup_guest.

@scott although refactor doesn't require tests, this refactor has no tests at all! 
The refactor won't have any problems though, because semantically is the same thing. However I think it would the better to open an issue to create tests specially for user::register since it's a very meaningful method.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -255,23 +255,13 @@ class User < ActiveRecord::Base
   end
 
   def self.register email, user_name
-    # this method is very similar to email_processor#create_user
-    # actually it was copyied from there.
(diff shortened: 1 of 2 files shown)


CGA1123:
#632

GPrimola:
Hi @scott ! Why this issue hasn't been merged yet?

scott:
To be honest I have not had a chance to fully review it yet.  I think there might be a couple other places were this could be used.

PR #642, opened 2017-06-08
update Persian translation

hadifarnoud:


Rynaro:
I see this PR falling in Travis, but saw in build logs, and nothing is really wrong with your PR. I just rerun the tests, and _All checks have passed_ status was returned by Travis.

The conflict tests appeared in this PR, because of an test that needs to guarantee order, and sometimes this order isn't guaranteed.

After all, this PR sounds good to me. I don't understand Persian, but I put my confidence on your work.
@scott , what do you think?

hadifarnoud:
can you please merge this?

Issue #631, opened 2017-05-19
Make ticket listings sortable by date (up and down)


scott:
Make the listing of tickets sortable by-

created_at asc
created_at desc
updated_at asc
updated_at desc
ticket number

This would add a new dropdown to the UI for sorting options.  There should also be a default sort order setting.

jcgalbano:
Is there no progress on this?

scott:
Nothing yet.

jcgalbano:
Okay I'll try to figure this out. Thanks!

scott:
Thanks @jcgalbano let us know if we can help.  This is may be a tricky one.

jcgalbano:
No prob :) btw does the updated_at asc and desc in this issue corresponds to the actual updated_at column of Topics or is it last_post_date? (which would be more intuitive to actually sort in this way)

jcgalbano:
#657

scott:
Just to clarify, this should be in the admin UI, `admin/topics/index` view

PR #657, opened 2017-06-29
#631 Make ticket listings sortable by date (up and down)

jcgalbano:
Since the app is already using turbo links I guess my implementation is valid by all means. Can you check up on this @scott? Any comments or suggestions will help a lot since I'm really new to open source thanks!


indagreen:
message not read? how many are using my account?

PR #691, opened 2017-08-27
Update french translation

tanji:
Big update of french translation. I corrected mistakes and translated all the missing strings.


scott:
Thanks for this!

PR #694, opened 2017-09-02
Fix jshint install

tanji:


CGA1123:
LGTM!

scott:
Thanks for fixing this! @tanji

PR #698, opened 2017-10-01
Remove byebug history

lidimayra:
Byebug history seems to be mistakenly committed. In general, this file is not versioned.

Delete file and add it to gitignore.


CGA1123:
This is good to merge. Travis failure related to #694

Issue #703, opened 2017-10-01
Add tagging to inbound webhook


scott:
It is possible to create new tickets via an inbound webhook.  When it was built, we forgot to add the ability to provide tags for the new ticket, so this would add the ability to submit tagging to the inbound webhook.  See `webhook/inbound_controller`.  This should be a really easy fix, but make sure you update the tests to include your changes.

query-string:
Hi @scott! Please take a look at #704 ;)

PR #704, opened 2017-10-02
Make it possible to send a list of tags over inbound webhook

query-string:
Quick fix + test for issue #703


scott:
Boom!  Thats awesome, thanks for the contribution.  Happy Hacktoberfesting :)

Issue #701, opened 2017-10-01
Add an icon if the ticket has any attachments


scott:
It would be awesome to have a little icon in `admin/topics/_topic.html.erb` that appears next to the ticket name if there is a file attached to the ticket.

esunder:
Would something like this be acceptable? How many different locations need the icon?

![image](https://user-images.githubusercontent.com/357754/31059921-dfc1f37a-a6ce-11e7-84ce-54e8c4c02506.png)

scott:
Yeah thats what i was thinking of.  Really the only place that needs it is the `_topic.html.erb` partial

esunder:
Ok, I will look at it. Thanks!

PR #705, opened 2017-10-02
Show attachment icon if tickets got attachment

firdausmaarof:
#701

### DIFF
--- a/app/models/topic.rb
+++ b/app/models/topic.rb
@@ -265,6 +265,19 @@ class Topic < ActiveRecord::Base
     end
   end
 
+  def check_attachments
+    attachments = false
+
+    self.posts.each do |post|
+      if post.attachments.present?
+        attachments = true
+        break
+      end
+    end
+
+    return attachments
+  end
+
   private
 
   def cache_user_name
--- a/app/views/topics/_topic.html.erb
+++ b/app/views/topics/_topic.html.erb
@@ -19,7 +19,14 @@
     </td>
   <% end %>
   <td>
-    <span class="more-important truncate-ellipsis"><%= link_to h(topic.name), topic_posts_path(topic) -%></span>
+    <span class="more-important truncate-ellipsis">
+      <a href="<%= topic_posts_path(topic) %>">
+        <%= h(topic.name) %> 
+        <% if topic.check_attachments %>
+          <i class="glyphicon glyphicon-paperclip"></i>          
+        <% end %>
+      </a>
+    </span>
     <span class="less-important"><%= t(:started_by, :username => topic.user.name.titleize) %></span>
   </td>
   <td>


app/models/topic.rb
```diff
@@ -265,6 +265,19 @@ def from_email_address
     end
   end
 
+  def check_attachments
```
CGA1123: I feel that maybe this should be renamed to `attachments?`

app/views/topics/_topic.html.erb
```diff
@@ -19,7 +19,14 @@
     </td>
   <% end %>
   <td>
-    <span class="more-important truncate-ellipsis"><%= link_to h(topic.name), topic_posts_path(topic) -%></span>
+    <span class="more-important truncate-ellipsis">
+      <a href="<%= topic_posts_path(topic) %>">
```
CGA1123: Any particular reason for favouring this over 
```
<%= link_to topic_posts_path(topic) do %>
...
<% end %>
```

CGA1123:
@firdausmaarof Looks great! Would it be possible for you to add tests for the new code you wrote?

app/models/topic.rb
```diff
@@ -265,6 +265,19 @@ def from_email_address
...
+    attachments = false
+
+    self.posts.each do |post|
```
CGA1123: Looking at the `Topic` model, it seems that an attachment may also be on the `Topic` itself (see [this](https://github.com/firdausmaarof/helpy/blob/2be47ccf3b947eb576e276a5d8251b018bc732f7/app/models/topic.rb#L42)), as well as its associated posts.

scott: I think this is true.  I can't remember how it works exactly though.  Good test cases would be 

- Create a new ticket with an attachment using the webform and verify the logo is shown
- Create a new ticket with an attachment by email and verify the logo is shown
- Create a new ticket without an attachment and reply to it with an attachment and verify the logo is shown

These tests already exist, so you can probably just re-use modify the code.

app/models/topic.rb
firdausmaarof: yeah. can, suddenly i feel its better to renamed it.

app/views/topics/_topic.html.erb
firdausmaarof: i didnt know we can wrap html elements in link_top helpers like that. cool, i'll change that.

scott:
@firdausmaarof can you please submit a couple test cases for this as I suggested above.  The PR cannot be accepted without this and its really only a couple lines of code since the relevant cases already exist :)

CGA1123:
In case you haven't done any unit testing before [here](https://semaphoreci.com/community/tutorials/getting-started-with-minitest) is a tutorial for the testing framework used by this project! (minitest).

Here are the [docs](http://docs.seattlerb.org/minitest/) for reference as well.

scott:
Not sure if he is going to come back and add the tests?

scott:
@firdausmaarof any chance you are going to return to this project and make any of the code changes you mentioned in the review?

Issue #702, opened 2017-10-01
Add priority to the Topic model


scott:
It would be useful to have a priority for tickets.  There is currently a priority for users/customers, but it would be good to add a separate priority for individual tickets as well.  The requirements for this:

- Priority values should be 'low','normal','high','very high' . normal should be the default.
- Add a new indexed string field to the `Topic` model for priority
- Provide a select field to set the priority from the `admin/topics/_new_topic` form
- Add the priority to the Topic model api entity and associated add and update endpoints in the Topic api.
- Add the ability to set the priority through the inbound webhook (`webhooks/inbound_controller.rb`

yoonwaiyan:
Hi, can I help out with this issue? I looked at the source code, would like to confirm the topic is represented as `message` in `inbound_controller.rb`?

CGA1123:
@yoonwaiyan Looking at the code for [`Webhook::InboundController#form`](https://github.com/helpyio/helpy/blob/5020e013efca901dad4c1cd89ee3dd7791e74ff9/app/controllers/webhook/inbound_controller.rb#L36) that seems to be the case.

scott:
Hey @Yoonwaiyan thanks for your interest.  Another person has already contacted me about this one, but I think you might be able to divide it up, since it touches a lot of places.  Let me see if he can join this thread and we can work it out so you each get a PR :)

yoonwaiyan:
@scott thanks, I agree and it's a good idea to divide this up :) FYI I just created a migration to add this column in my fork, I can create a PR first or I can take the first 2-3 subtasks as mentioned.

scott:
Yeah go ahead and do a PR for the migration and then you can work on the subtasks.

jeffdill2:
Hi @yoonwaiyan, dividing it up sounds good to me. 😄

@scott , did you have any thoughts about the best way to break this up?

scott:
I think one person could take the UI changes, and webhook, the other the api?  overall its not a huge amount of work

yoonwaiyan:
I would like to take the Grape API/entity, what do you think @jeffdill2 ?

jeffdill2:
@yoonwaiyan go for it ☺️

jeffdill2:
@yoonwaiyan you can actually take the whole thing. I'm getting pulled into another project. Sorry!

PR #706, opened 2017-10-03
Add priority column to Topic

yoonwaiyan:
For issue #702. Only added migration and run `annotate` to populate schema information into models.


CGA1123:
Would an [`enum`](https://apidock.com/rails/ActiveRecord/Enum) not be more appropriate for this?

yoonwaiyan:
I followed the priority in User which is using string to store values but enum is a good idea. Adjusted to use enum(integer) instead of string. Currently it's set to 0,1,2,3 respectively and use 1 as the default(normal).

app/models/topic.rb
```diff
@@ -86,7 +86,7 @@ class Topic < ActiveRecord::Base
   validates :name, presence: true, length: { maximum: 255 }
   # validates :user_id, presence: true
 
-  enum priority: { normal: 0, high: 1, very_high: 2 }
+  enum priority: { low: 0, normal: 1, high: 2, very_high: 3 }
```
jeffdill2: FWIW, you could just do `enum priority: [:low, :normal, :high, :very_high]` here. Rails will implicitly use 0 - 3 when storing the integer in the database. Definitely nothing wrong with declaring it explicitly though. :smile:

yoonwaiyan: I have experience of defining enum differently (increments of 10 instead of 1) in projects, used to define enum explicitly 😅 but let's see if it's acceptable this way.

jeffdill2: Right on 👍

scott: I think its better to delare specific integers in case you ever want to add some additional level you can.  See this blog post https://bendyworks.com/blog/rails-enum-sharp-knife

yoonwaiyan: @scott So is there anything I can improve from the current line? It's being explicitly defined for now and it's possible to put in other values in between the values

scott:
@yoonwaiyan do you want to work on the rest of this task?  @jeffdill2 said he will not have time to do anything else on it.

yoonwaiyan:
@scott I can take the rest if nobody is interested. So this pull request will be remained open until I finish all tasks?

scott:
Yeah I can leave it open or you can make another with the other work if you want.

yoonwaiyan:
I did all tasks that are mentioned in the issue, do update me if I missed out anything or any specs. I have  merged the branch from master, seems like there's one failing specs in master.

scott:
Thanks again for doing this feature @yoonwaiyan sorry it took so long to get it merged, I had some personal/family stuff come up this fall and I was unable to do much then.

PR #707, opened 2017-10-03
Model fixes

swatijadhav:
This PR covers following:

- [x] Update model code, use simple conditions
- [x] Use factories to create test records
- [x] Update test cases for models

Changes are still in progress, raising PR to know the comments or any suggestions on the changes.

Thanks

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -110,13 +110,11 @@ class User < ActiveRecord::Base
   scope :admins, -> { where('admin = ? OR role = ?',true,'admin').order('name asc') }
   scope :agents, -> { where('admin = ? OR role = ? OR role = ?',true,'admin','agent').order('name asc') }
   scope :team, -> { where('admin = ? OR role = ? OR role = ? OR role = ?',true,'admin','agent','editor').order('name asc') }
-  scope :active, -> { where('active = ?', true)}
-  scope :by_role, -> (role) { where('role = ?', role) }
+  scope :active, -> { where(active: true)}
+  scope :by_role, -> (role) { where(role: role) }
 
   def set_role_on_invitation_accept
-    if self.role.nil?
-      self.role = "agent"
-    end
+    self.role = self.role.presence || "agent"
     self.active = true
     self.save
   end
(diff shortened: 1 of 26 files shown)


app/models/user.rb
```diff
@@ -110,13 +110,11 @@ class User < ActiveRecord::Base
...
-  scope :active, -> { where('active = ?', true)}
-  scope :by_role, -> (role) { where('role = ?', role) }
+  scope :active, -> { where(active: true)}
+  scope :by_role, -> (role) { where(role: role) }
 
   def set_role_on_invitation_accept
-    if self.role.nil?
-      self.role = "agent"
-    end
+    self.role = self.role.presence || "agent"
```
CGA1123: `self.role ||= "agent"`, is even cleaner imo.

swatijadhav: the reason to use `self.role.presence` as it will also exclude blank string.

     "" || "agent"
     => ""

CGA1123: In theory the roles are predefined and should only ever be one of `[admin, agent, editor, user]`. 

Although I don't see any enforcement of this within the model.
Maybe that should be looked into, good point on the blank string case though.

swatijadhav:
@CGA1123 Changes are now ready for review. Can you please verify and merge. Thanks.

CGA1123:
LGTM!
I don't have the permission to merge unfortunately. Will have to wait for @scott to approve and merge.

@scott, is there a project wide push toward replacing fixtures with `FactoryGirl`?

app/models/user.rb
scott: I would leave it unrestricted in case someone wants to add their own values.

app/models/user.rb
```diff
@@ -130,15 +128,15 @@ def enable_notifications_for_admin
...
   end
 
   def self.notifiable_on_private
-    User.agents.where(notify_on_private: true).reorder('id asc')
+    self.agents.where(notify_on_private: true).reorder('id asc')
   end
 
   def self.notifiable_on_reply
-    User.agents.where(notify_on_reply: true).reorder('id asc')
+    self.agents.where(notify_on_reply: true).reorder('id asc')
```
scott: Is it in the style guide to prefer `self` here?  Also, why not just shorten it to 
```
def self.notifiable_on_public
   agents.where(notify_on_public: true).reorder('id asc')
end
```

swatijadhav:
Changes are done as per the suggestion. Thanks @scott

PR #708, opened 2017-10-03
Extract onboarding

scott:
Extracts the onboarding into a vendored engine, just to clean up the core codebase slightly.


CGA1123:
Have you thought of fully extracting this into its own repo?

Might as well go all the way!

scott:
closes #700

Issue #656, opened 2017-06-28
Internal KB test suite


scott:
There are currently no tests for the internal knowledge base developed by @Rynaro  This ticket is so we don't forget about it.

PR #718, opened 2017-10-18
Controller tests for internal knowledge base

andrewstucki:
Related to https://github.com/helpyio/helpy/issues/656

It's not exactly a full test suite, but it's basic controller tests that appear to have actually caught a bug with the controller-based redirects (due to nested controllers).


scott:
Hey @andrewstucki thanks so much for this!

PR #721, opened 2017-10-19
Enabling Rubocop

### LINKED ISSUE #709: Rubocop Disabled By Default
While the project does have `rubocop` enabled, and Travis it. The config options `DisabledByDefault: true` is set.

This means all rules are by default **not applied**. (Apart from `Rails` Cops and the `Style/AndOr` cop which have been explicitly enabled.)

I think it would be a good idea to enable the rules by default, and have a conversation about what rules to explicitly turn off/modify.

CGA1123:
~Related to #709~ Closes #709 

This PR enables `rubocop` by default! As well as disabling/configuring some specific cops that are annoying/would cause a huge pain on this codebase!

~I also run the codebase through rubocops autofix feature, which fixes a lot of the simple errors!~
EDIT: auto fixing isn't really usable as we are still supportting older ruby version - autofix doesnt take that into account, creating breaking changes in the codebase.


CGA1123:
What are your thoughts on this @scott ?

PR #739, opened 2017-12-13
define charset,search engine and external css

keangkai:
In the future css increase many, search engine and define charset = UTF-8


scott:
This is a file that ships with Rails and has not been customized.  We are moving to rails 5 which may change this anyways, so I am closing for now.  Feel free to reopen if you can make a stronger justification for it once the Rails 5 branch has been merged in.

PR #750, opened 2017-12-29
Create devise/ms.yml

brahim-r:
File in Malay language


scott:
Hey @brahim-r thanks so much for providing these locales!  I will get the AR in too, I just need to figure out why the tests are not passing and it seems Travis is not working right now.

PR #768, opened 2018-01-23
Copy CC emails to ticket

### LINKED ISSUE #767: When creating a ticket via email, any CC'd addresses should be added to the CC of the thread
Currently you can add a CC to a ticket manually.  If a ticket is received via email, those CC'd on the email should be added to the ticket CC as well.

CGA1123:
Closes #767


patthewebrat:
Hey guys,
Thanks for merging in this fix, I've been testing it and it's mostly working ok.
One thing I have noticed however is that even though the cc'ed email address appears in the thread 
it then disappears from the thread again unless you expand the cc panel on the reply.
Thanks,
Pat

PR #788, opened 2018-02-22
Refactor change user functionality to remove duplicate html ids

CGA1123:


scott:
This looks about how i would have done it too.  I noticed two problems on reviewing it though:

1. If you click from the ticket listing into a ticket, then try to change the author, the modal shows, then disappears on its own.
2. I added another link to change author on the ticket options menu (shown first in the video) and that is not working at all, probably link needs to be changed/add to the js.

change_author

scott:
Any updates on this @CGA1123

PR #799, opened 2018-02-27
Add first to post

scott:


app/models/post.rb
```diff
@@ -125,6 +125,10 @@ def importing?
     self.importing || false
   end
 
+  def first?
+    self.topic.posts.first == self
```
CGA1123: Would checking `kind == 'first'` here achieve the same and be less database intensive?

scott: Thats how it was previously (when you wrote it.)  However we are relying on the value being 'first' which may not always be the case.  We have a case now where we are using a different value but it is still the first post in the series.  

Although it is more expensive to do it with the lookup, it enforces integrity.

PR #741, opened 2017-12-14
Imap support

j0ack:
I've done the revert work mentioned here #221 to be allow to use our own SMTP server.

Since this work seems old let me know if this PR is outdated and what I can do to be able to merge this functionality.

PR #810, opened 2018-03-29
IMAP working

### LINKED ISSUE #221: imap support
ability to fetch inbound email using imap(s) would be nice

scott:
This seems to be working now, still need to verify griddler still works, get tests to pass, etc.

closes #741 
closes #221 
closes #179
#148


hadifarnoud:
this essentially solves my issue with having custom griddle.

hadifarnoud:
one question, if the email address is a user of Helpy, then it gets an email for every new ticket. that creates a loop!

do you have any plan to prevent this?

scott:
@hadifarnoud somehow I missed this- yes we fixed this in the new version that just merged in yesterday.  However before you upgrade, I expect there may be issues with RTL in the changes we made.

PR #837, opened 2018-07-08
Update fa.yml

hadifarnoud:
updated Persian translation with all the new strings.
fixed a few typos and bad translations


scott:
I am not sure why its reporting a conflict? Can you resolve?

PR #846, opened 2018-07-13
Android Support

trevormerritt:
Adds QR code, API Endpoints.

Please Squash on merge.  1 is to fix a pair of vulnerable gems.  1 is to revert to default config from my dev environment.


scott:
@trevormerritt  Remind me, what is the QR code stuff for?  The API endpoints look good, I would probably cherry-pick them into a new PR.

trevormerritt:
When adding an Android client, typing a long API key sucks. The QR code has the URL of the API and the API key. 
The client I've been writing does this now.

scott:
Closing due to inactivity. Whatever happened with this @trevormerritt ?

trevormerritt:
The QR code is so the Android client can use that to enter the API key
rather then a long string of random characters.

PR #853, opened 2018-07-16
Updated Uglifier

### LINKED ISSUE #852: Summernote editor broken
Summernote editor is broken and not instantiating correctly on the reply form/new ticket form, ONLY IN PRODUCTION MODE.  A short term workaround is to force the version back of the `summernote-rails` gem.  I know `0.8.3.0` works.

scott:
Fixes #852 Updating uglifier resolved the bug in the compilied JS.


scott:
Appears to have merged but GH did not update because of issues

PR #877, opened 2018-08-21
Use Ruby function 'pluralize' correctly with locale

ottok:
One cannot pluralize with English rules for every language. For example in Finnish all words ended up with an extra 's' at the end, which is nonsense. This fixes it so that pluralization is always done correctly: if Ruby knows the rules for the language, the word is pluralized, and if Ruby does not, the word will not be mangled randomly.

Before:
screenshot from 2018-08-21 12-40-24
After:
screenshot from 2018-08-21 13-13-38

After (English is still correct):
screenshot from 2018-08-21 13-13-52

### DIFF
--- a/app/views/admin/users/_user_filter.html.erb
+++ b/app/views/admin/users/_user_filter.html.erb
@@ -4,9 +4,9 @@
 </div>
 
 <ul class="list-unstyled nav-items">
-  <%= render partial: 'admin/users/user_nav_item', locals: { role: 'user', label: t(:user_role).pluralize } %>
-  <%= render partial: 'admin/users/user_nav_item', locals: { role: 'team', label: t(:team) } %>
-  <%= render partial: 'admin/users/user_nav_item', locals: { role: 'admin', label: t(:admin_role).pluralize } %>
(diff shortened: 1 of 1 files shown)


ottok:
See docs: https://apidock.com/rails/String/pluralize

ottok:
PIng, any updates to this?

scott:
Hey @ottok finally had a chance to review this and consider some alternative solutions to this problem.  It exists not just for `fi` but other locales and gets worse with more non-latin charsets (ie. KO, JP, RU etc.)  The proposed approach fixes the bug for some languages and breaks it for others (like FR, ES, PT) which generally always add "s" to indicate plurality. 

The issue is that the pluralize backend in rails only includes rules for English, so it is applying this rule to all locales including those that use a different concept for pluralization (as you mentioned).  One solution is to write/add additional language specific rules (called "inflectors") for each locale.  I looked for a central repository for this figuring someone had already done it, but didn't find one and don't want to go to the trouble of trying to do this for 25+ languages.

So... I think the best solution is actually to change the UI in Helpy to only show singular representations of each role, which is probably more gramatically correct anyhow.  I will open a separate PR for this which will remove the pluralization completely and will close this issue once I merge it in.  Thanks so much for taking the time to identify and investigate this issue!

PR #879, opened 2018-08-21
Add a link to edit page of docs

elguitar:
### What
If the user is editor or up, a link will appear on doc pages. The link will lead to the edit page of the corresponding doc.

### Why
This provides a better workflow for editors to correct spelling mistakes etc.

Please provide feedback and/or ask questions about this. I am more than happy to correct the code and answer questions if needed.

### EDIT
Here's the screenshot @ottok proposed:
2018-08-22-095155_1179x408_scrot

### DIFF
--- a/app/views/docs/show.html.erb
+++ b/app/views/docs/show.html.erb
@@ -27,7 +27,13 @@
   background-color: transparent;
 }
 </style>
-
+<% if current_user && current_user.is_editor? %>
+<div class="row edit-row">
+  <div id="doc-edit" class="text-right">
+    <div class="more-important meta-header"><%= link_to I18n.t("edit", default: "Edit"), edit_admin_category_doc_path(@doc.category.id, @doc.id, lang: I18n.locale) %></div>
+  </div>
+</div>
+<% end %>
 <div class="row content-row">
   <div class="col-md-3 hidden-xs hidden-sm">
     <div id="written-by" class="doc-meta">


ottok:
As this is an UI change, please add a screenshot :)

Pyo25:
Instead of adding a new row, what do you think about adding a small "edit" (or an icon) next to the title?

Would be also great to add this to the `singular` theme: https://github.com/helpyio/helpy/blob/master/app/themes/singular/views/docs/show.html.erb

scott:
I think this is a good idea, but would better be implemented as some sort of javascript toolbar that would only appear if the user is logged in as an editor/admin/agent.  That way it could be included into the header or footer as a simple js file, and should work no matter the theme used.  Thoughts?

mikkoaf:
There is also the option to utilize the summernote API to allow admins to edit the docs https://summernote.org/examples/#click-to-edit

ottok:
This is a really annoying problem and I wish come way to fix it would be applied soon. Consider this use case: I find a typo on https://help.seravo.com/fi/knowledgebase/15/docs/98-sivuston-eheyden-ja-turvallisuuden-tarkistus. I would like to just click "Edit" and fix it, as I am logged in as an admin. However, at the moment I need to click with my mouse 6 times before I get to the edit view: `Menu > Admin > Helpcenter > Contents > Category name > Article name`.

Pyo25:
@mikkoaf I would prefer the solution suggested by @scott : click-to-edit is handy but it's acceptable only for text fields (what if you want to edit tags, meta title, etc.?). Moreover, it adds quite a lot of logic to the frontend.

@ottok Have a look at @scott 's comment and feel free to open a PR. 
A workaround is to create your own theme and add the edit link where you want.

scott:
I think another way is to accomplish this is to use a custom js script inserted via the settings > design > footerjs to add the link to the page.  You can inspect the sidebar menu for a link to "admin" to see if the currently logged in user is an admin, and only then show the edit link.  This would not require any changes to the Helpy codebase and would give you the direct link to the editing page.  The only drawback is lack of localization on the link text.

olliebennett:
TL;DR - I think the solution implemented here is the best way to go (UI placement aside).

I'm not sure about;

> [a JavaScript solution] should work no matter the theme used

I can see the appeal of not needing to add this link to every theme, but by adding it via JavaScript do you not have the problem that the link placement, icon, text etc. can no longer be modified per-theme? You'd also rely on certain HTML structure for this to work (so your theme would need to implement that layout structure).

I think this is a good candidate to implement in Rails code instead of JavaScript, and adding the few simple lines (see below) to each theme shouldn't be too difficult.

In the interest of avoiding multiple implementations per-theme, we could also refactor the permissions logic into something like a `@secondary_links` variable which could be `[]` for normal visitors, but be `[{ key: :edit, title: 'Edit', link: edit_url }]` for admins. In future we could add more such links - perhaps "permalink" or "share". This reduces the duplication/logic in the themes to something more like;

```ruby
@secondary_links.each do |sl|
  # insert new row or inline icon/text/etc.
end
```

As for how this should appear in the UI for the default theme, I also vote for avoiding a whole new row to keep things compact, but that's just a UI decision that could of course be made per-theme.

Issue #828, opened 2018-06-12
Internal KB articles are shown multiple times


scott:
This is a rare bug where a translated category is duplicated in the UI, only if the user is viewing the admin in the non default locale.  To reproduce- create a new internal category and doc, then edit the category to create an alternate locale version.  Now view the internal docs using the same alternate locale and you will see the doc repeated.

dhoangk07:
Hi @scott, I've tried to reproduce the bug as your description but it doesn't happen on my end. Could you give me more details of steps or record the video? It's definitely helpful. Thanks

scott:
Hey @dhoangk07 let me do that.

scott:
Here take a look at this- https://www.screencast.com/t/CTz6F9e0W0 . If you don't have it, you might grap the "Quick Language Switcher" add on for Chrome, makes switching between locales much easier.

dhoangk07:
Hi @scott I have PR to fix the issue. Please take a look https://github.com/helpyio/helpy/pull/904

PR #904, opened 2018-09-20
Use current locale to filter category

dhoangk07:
## Context
The categories with multiple locales appears multiple times because the query does not filter with current locale

## Changes
Use scope to filter category with current locale

## References
https://github.com/helpyio/helpy/issues/828

### DIFF
--- a/app/controllers/admin/internal_categories_controller.rb
+++ b/app/controllers/admin/internal_categories_controller.rb
@@ -1,7 +1,7 @@
 class Admin::InternalCategoriesController < Admin::BaseController
 
   def index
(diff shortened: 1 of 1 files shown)


app/controllers/admin/internal_categories_controller.rb
```diff
@@ -1,7 +1,7 @@
 class Admin::InternalCategoriesController < Admin::BaseController
 
   def index
-    @categories = Category.internally.without_system_resource.ordered
+    @categories = Category.internally.without_system_resource.with_translations(I18n.locale).ordered
```
scott: Good catch!

Issue #821, opened 2018-05-22
Ticket permission bug (rare)


scott:
There is a bug where a private ticket can still be in the public search index if you toggle it and reply.

PR #905, opened 2018-09-20
Dev/821 ticket permission bug

dhoangk07:
## Context
  PG search does not update search document for the topic when switch from public to private because of `update_all` method which skips callbacks

## Change
  Call `update_pg_search_document` function on changes topic.

## Reference
  https://github.com/helpyio/helpy/issues/821


scott:
@dhoangk07 Oops looks like this created a conflict because I merged one of your other PRs first.

PR #934, opened 2018-10-17
Remove code duplication in application_helper.rb

nikobozi:
In `application_helper.rb` there are 2 methods `tag_listing` and `doc_tag_listing` which execute identical code. Since both are being used I left them there, but removed the duplication

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -90,16 +90,12 @@ module ApplicationHelper
   def tag_listing(tags, tagging_type = "message")
     return unless tagging_type == "doc"
     tags.each do |tag|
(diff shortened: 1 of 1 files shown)


olliebennett:
Nice clean-up. I wonder if it's worth using `alias_method` (which should achieve the same thing), or else replacing any calls to `doc_tag_listing` with `tag_listing` elsewhere in code directly, and remove the method?

👍 from me to merge this anyway if you don't want to take it further! :)

nikobozi:
> Nice clean-up. I wonder if it's worth using `alias_method` (which should achieve the same thing), or else replacing any calls to `doc_tag_listing` with `tag_listing` elsewhere in code directly, and remove the method?
> 
> +1 from me to merge this anyway if you don't want to take it further! :)

I like the idea with the alias method! I didn't want to replace all the calls in case anything breaks, but could try that as well

PR #951, opened 2018-10-22
Improve Docker support

ypcs:
This PR improves Docker support. As a result, build time decreases a lot, it's easier to use Docker for development, and less network traffic.

 - use local clone for building the image (instead of cloning remote repository)
 - use local Dockerfile to build Helpy instead of pulling remote `helpy/helpy` in `docker-compose` to make development much easier
 - copy only necessary parts in early steps of the build to improve build speed, and only in the end copy rest of the content
 - merge multiple configurations to `docker/.env{,.sample}` to make configuration easier
 - move PostgreSQL data directory to temporary volume
 - update documentation to match current status
 - etc.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,8 +8,9 @@ ENV RAILS_ENV=production \
 RUN apt-get update \
   && apt-get upgrade -y \
   && apt-get install -y nodejs postgresql-client imagemagick --no-install-recommends \
-  && rm -rf /var/lib/apt/lists/* \
-  && useradd --no-create-home $HELPY_USER \
+  && rm -rf /var/lib/apt/lists/*
+
+RUN useradd --no-create-home $HELPY_USER \
   && mkdir -p $HELPY_HOME \
   && chown -R $HELPY_USER:$HELPY_USER $HELPY_HOME /usr/local/lib/ruby /usr/local/bundle
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,6 @@
 FROM ruby:2.4
 
-ENV HELPY_VERSION=master \
-    RAILS_ENV=production \
+ENV RAILS_ENV=production \
     HELPY_HOME=/helpy \
     HELPY_USER=helpyuser \
     HELPY_SLACK_INTEGRATION_ENABLED=true
--- a/Dockerfile
+++ b/Dockerfile
@@ -49,6 +49,5 @@ RUN chown -R $HELPY_USER $HELPY_HOME
 USER $HELPY_USER
 
 COPY docker/database.yml $HELPY_HOME/config/database.yml
-COPY docker/run.sh $HELPY_HOME/run.sh
 
-CMD ["./run.sh"]
+CMD ["/bin/bash", "/helpy/docker/run.sh"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,9 +15,12 @@ RUN apt-get update \
 
 WORKDIR $HELPY_HOME
 
+COPY Gemfile Gemfile.lock $HELPY_HOME/
+COPY vendor $HELPY_HOME/vendor
+RUN chown -R $HELPY_USER $HELPY_HOME
+
 USER $HELPY_USER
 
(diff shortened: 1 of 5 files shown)


scott:
Thanks for this work @ypcs 👍  One question... I saw you moved the location of the postgres database - is this still retained between updates when running in production?  How will this affect production installations that are already running Postgres in the prior location?

ypcs:
Data is retained between updates *after switching to this new volume*, it's just stored in Docker volume directory instead of `$(pwd)/postgres`. You can list current Docker volumes with `docker volume ls`. Default name seems to be `<name of current directory>_postgres`, ie. for me this looks like

```shell
$ docker volume ls
DRIVER              VOLUME NAME
local               helpy_postgres
local               helpy_rails-assets
```

And if I'm interested in more details, `docker volume inspect <volume>` comes to help,

```shell
$ docker volume inspect helpy_postgres
[
    {
        "CreatedAt": "2018-10-22T20:18:17+03:00",
        "Driver": "local",
        "Labels": null,
        "Mountpoint": "/var/lib/docker/volumes/helpy_postgres/_data",
        "Name": "helpy_postgres",
        "Options": null,
        "Scope": "local"
    }
]
```

These volumes persist, until removed by user, eg. by running `docker-compose rm`. Tested by running

```
# do the initial build & run container
docker-compose build
docker-compose up -d

# make changes to helpy code
echo "#mychange" >>README.md

# rebuild the image & relaunch
docker-compose build
docker-compose up -d

# now running updated code, and contents in postgres volume were kept
# also by looking data via `docker volume inspect <volume>` timestamps weren't changed
```

**HOWEVER** current installations (created with volume `./postgres:...`) will not switch to use this new path without manual step, like. `sudo cp -a postgres/* /var/lib/docker/volumes/helpy_postgres/_data/`.

I'm fine if that single commit (6f86519) gets dropped, workflow will be enhanced anyways.

ypcs:
@scott is the behaviour described above OK, or should I remove changes to Postgres data directory? (And perhaps create separate PR from it.)

scott:
@ypcs I think breaking that into its own PR would make sense, since it does represent a breaking change for anyone using the current compose layout.

ypcs:
I dropped commit containing changes to Postgres data directory. Will create separate pull request from it.

ottok:
Having a solid Dockerfile in the project is important for attracting new contributors. It is so much simpler to test and develop Helpy if a simple `git clone`+ `docker build` + `docker run` is enough to get a dev environment running. This PR is an important step towards that. Please @scott consider merging this soon. Thanks!

scott:
hey @ypcs this PR is generating a bunch of errors, which the previous one did not.  Can you take a look so I can get it merged in?

ypcs:
It seems that renaming `scss-lint` to `scss_lint` (as suggested by output from `bundle install`) broke the tests, so I removed that change. It was anyways out-of-scope for this PR.

ypcs:
@scott I've rebased this to match current master, and tests seem to have passed since.

Dockerfile
```diff
@@ -1,43 +1,49 @@
...
 # manually create the /helpy/public/assets folder and give the helpy user rights to it
 # this ensures that helpy can write precompiled assets to it
-RUN mkdir -p $HELPY_HOME/public/assets && chown $HELPY_USER $HELPY_HOME/public/assets
+RUN mkdir -p $HELPY_HOME/public/assets \
+    && chown $HELPY_USER $HELPY_HOME/public/assets
 
 VOLUME $HELPY_HOME/public
 
+USER root
+COPY . $HELPY_HOME/
```
scott: Hey @ypcs I merged this work in, but wouldn't this line overwrite the helpy_slack gem addition on line 31?

ypcs: Yes, it in fact looks like that. Gem should however get installed, as `bundle install` gets executed before copying contents (including Gemfile) to `$HELPY_HOME`.

Should that helpy_slack part be rewritten so that module gets always installed (= is always defined in Gemfile, no need to sed it in), and at runtime app just checks if Slack integration should be activated?

Issue #947, opened 2018-10-19
Checkbox location on sign in is off


msbt:
As mentioned on slack: The checkbox location on login is off, e.g. /en/users/sign_in (has a min-height: 60; solution would be #user_remember_me { min-height: auto; }

scott:
@msbt I looked at this, and can't seem to see where/how it is off.  I also reviewed the proposed solution by @dhoangk07 and I must be looking at the wrong thing because I can't see the original issue or any difference in the PR.   Can you post a screen cap and clarify which browser and theme you are seeing this on?

msbt:
Ok that's odd, I've just tested chrome, Firefox and Edge on Win10 and they all show the same error. When I use OSX both, Safari and chrome show it correctly.

![helpy](https://user-images.githubusercontent.com/8767096/47441693-fb971a80-d7b0-11e8-984f-ded6b281b041.png)

scott:
Well there you go.  I am using a Mac and don't see the problem.  SMH.

dhoangk07:
Hi @scott, @msbt 

Does the PR https://github.com/helpyio/helpy/pull/952 look good?

scott:
I agree on making this change more specific.  Also, you should check other themes as they may have the same issue.  You can add ?theme=light ?theme=flat ?theme=helpy to see this.

msbt:
As far as I can tell they're looking as expected because they don't have that much padding

![helpy3](https://user-images.githubusercontent.com/8767096/47448600-b5e24e00-d7c0-11e8-8e09-66e2458e5613.png)

dhoangk07:
Hi @scott , @msbt.

Please refer to attached file for Firefox on Mac after adjust `min-height` from `60px` to `auto` which above is original one.

<img width="1440" alt="screen shot 2018-10-25 at 9 20 58 am" src="https://user-images.githubusercontent.com/32945097/47472280-5031a880-d838-11e8-8f73-4cab6baad0a2.png">
<img width="1440" alt="screen shot 2018-10-25 at 9 21 22 am" src="https://user-images.githubusercontent.com/32945097/47472281-50ca3f00-d838-11e8-9f65-41e5e1c91c85.png">

msbt:
Thanks for your efforts @dhoangk07 
That's good to know, Firefox on OSX renders the same as it does on Windows. But as you can see, with your solution the input fields are also smaller which is not intended I guess.

scott:
[...]

PR #952, opened 2018-10-22
Change min-height of remember me checkbox

dhoangk07:
## Change 

Change min-height remember checkbox

https://github.com/helpyio/helpy/issues/947


scott:
I could not see any actual difference when I tested this change?

dhoangk07:
Close PR due to https://github.com/helpyio/helpy/pull/980/files

Issue #854, opened 2018-07-17
Validate URL in settings


scott:
Currently the site URL setting will allow anything.  If the user mistakenly enters something that is not a URL, parts of Helpy will not work correctly.

scott:
#873

ozmar-11:
can I take this one?

scott:
Hi @ozmar-11 that would be great! You'll notice that some work was already done in #986 so if you would review my notes there that might be helpful.

PR #986, opened 2018-10-27
Add Site URL validation to avoid unexpected errors

dsantosmerino:
Solves #854

With this new validation, we'll prevent the application to start when an admin has introduced an invalid URL on the Settings config file.


scott:
Thanks for your contribution @dsantosmerino! Instead of simply raising an error, how about using HTML 5 validations on the client side code to enforce the correct/desired formatting?

PR #988, opened 2018-10-29
Dev/754 attachments not showing up in e mail

dhoangk07:


scott:
Hey @dhoangk07 thanks for taking a look at this issue!  I like what you have done and it looks like it will work for tickets with attachments that are created through the UI.  However, most attachments arrive by email and are saved as the email is processed.  In this case, it will still not work because the file upload is handled by a job- which means the only real way to make sure the attachments are attached is to either wait to send out the notification, or add some kind of callback. Thoughts?

dhoangk07:
Hey @scott .
I am so sorry for reply late.
Regarding above ticket, please help me how to upload attachment arrive by email. I want to reproduce on my own machine. Could you give me more details of steps or record the video? It's definitely helpful. Thanks

scott:
@dhoangk07 thanks- working on this...

PR #1004, opened 2018-11-01
Add a search field in article and category view in Flat and Light themes.

cr0vy:
This PR adds a search field to the Flat and Light themes in article and category views.


scott:
Hey @cr0vy thanks for adding this PR. I generally don't take PRs that materially change an existing theme, but you are welcome to incorporate this into a new theme, or you may be able to accomplish this with javascript added through the settings > design.

ottok:
If a new theme is created, will that be incorporated into the hosted edition of Helpy and available to Helpy customers? As a hosted edition customer, I am specifically interested in development that helps the customer experience of hosted edition users get better.

scott:
Yes, that is what I was thinking.  If there was a new theme (say light2) or something like that we would incorporate it into both the open source and hosted versions.

PR #1027, opened 2018-11-15
Update Finnish translation

cr0vy:
This PR updates Finnish the translations under config/locales/devise .


scott:
Thanks @cr0vy I merged this in, but you might also look at sending a PR to https://github.com/tigrish/devise-i18n if you haven't yet!

PR #1059, opened 2018-12-14
New theme for Helpy by Seravo: Nordic

cr0vy:
We currently use Flat theme on our hosted Helpy site. We’ve had a few issues with the Flat theme and need more features than the theme can currently provide. Those issues and features originate from the theme in use, so we started working on a new theme. 
Nordic theme is fork of Flat theme. We have fixed some small issues and added few basic features we needed. This PR contains the base of theme and we will continue to improve it in the future.

What Nordic theme improves:
**Search field on every page in Helpy**
Gives customers an easy way to search for an another article, for example in different articles and on the category view.

**Improve the accessibility of the language selection**
Changing the language can now be done from any page without leaving the current page. 
screenshot 2018-12-13 15 27 04

**Cleaner URLs**
The URLs have been shortened, for example from example.com/en/knowledgebase/1-example-category/docs/1-example-article to example.com/en/docs/1-example-article 
screenshot 2018-12-13 15 28 04

**Editing article via button**
Admins can now go directly to editing from the the article page.
screenshot 2018-12-13 15 27 28

**Legibility**
Optimize typography to make support documentation as easy to read as possible.
screenshot 2018-12-13 15 26 51


scott:
This looks like a great contribution @cr0vy ! I will pull this and make any suggestions, as well as look at the Brakeman XSS warnings...

scott:
#879 
#1004

elguitar:
I pushed the brakeman ignores which are required to get past false-positive from brakeman in 34ff1ad.
Should it be corrected to the first commit and then build this branch again or is it fine like this?

scott:
It's fine like this!

ottok:
Thanks @scott !

PR #1062, opened 2018-12-26
Enable click-outside-to-close with keyboard shortcuts modal

elguitar:
The modal was defined to be static and there was no button to close the modal. Therefore the only way to close it was by Esc-key.

With this PR, clicking outside of the modal closes the modal.

If this is not the desired way of closing the modal, I can implement something else that would make the UX better. For example a big X in the corner.


scott:
Hi @elguitar this is a nice PR and a good idea for a UI improvement.  In addition to the click outside modal, can you also implement an X in the corner to close?

scott:
@elguitar I am gonna just merge this, please send a follow up when you get the UI part done.

PR #1126, opened 2019-01-30
Extended Ru translation 

lawpin:


scott:
Thanks @lawpin 🙌

lawpin:
@scott, more parameters needed in yml.

categories
edit_note
edit_ticket
new_ticket
save_category
users

PR #1166, opened 2019-03-02
Add onboarding flow on first browse rather than first login

scott:


scott:
Moved into separate repo, updated there

Issue #77, opened 2016-03-07
Change google fonts to use https to avoid "mixed content" warning


pepjo:
When using helpy over https one gets the following message on chrome:
````
Mixed Content: The page at 'https://example.io/ca' was loaded over HTTPS, but requested an insecure stylesheet 'http://fonts.googleapis.com/css?family=Droid+Sans:400,700|Droid+Serif:400,700|Lobster'. This request has been blocked; the content must be served over HTTPS.
````

I'll send a pull request to fix it :smile:

PR #1178, opened 2019-03-11
Fixed mixed content issue in the widget

karser:
When using widget over https I'm getting a mixed content issue, looks like similar to #77


scott:
Thanks @karser 👍

PR #1215, opened 2019-03-29
Saml

pitbulk:
This adds SAML support to Helpy.

btw, I updated the info available at: https://github.com/helpyio/helpy/wiki/SSO--Configuring-OAUTH

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -59,6 +59,7 @@ gem 'omniauth-github'
 gem 'omniauth-twitter'
 gem 'omniauth-google-oauth2'
 gem 'omniauth-facebook'
+gem 'omniauth-saml'
 
 # i18n gems
 gem 'rails-i18n', '~> 4.0.9'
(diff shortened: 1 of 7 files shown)


pitbulk:
There is a pending improvement that need to be coded, I hope is not hard for you.

At the [sessions_controller.rb](https://github.com/helpyio/helpy/blob/93e8e36d9c1b1fb8861203ba5f432ec0aac8c6a9/app/controllers/sessions_controller.rb
) file, in the after_sign_out_path_for method, the should look like
```
def after_sign_out_path_for(_)
    idp_slo_url_defined = <check that the idp_slo_target_url setting is defined or not>
    if session.key?('saml_uid') && session["saml_uid"] && idp_slo_url_defined
      session.delete("saml_uid")
      user_saml_omniauth_authorize_path + "/spslo"
    else
      super
    end
  end
```

Issue #1238, opened 2019-04-16
no way to search for posts by content from API


trevormerritt:
the API is missing a way to search for posts.

PR #1239, opened 2019-04-16
Adds search for Posts by Content

trevormerritt:
Addresses #1238


trevormerritt:
Wow.  pg_search_scope is a slick gem.  I'm reworking my submission.  Thanks for the pointers.

trevormerritt:
See PR 2191

Issue #961, opened 2018-10-22
Add honeypot to new ticket form


scott:
Helpy gives the option to add a recaptcha to the new ticket form (`topics/new`).  If no re-captcha is configured, we should still add some kind of honeypot anti-spam field to the new ticket form.  Obviously this could be hidden if the re-captcha is displayed.

Brand-Temp:
Hi! I'd like to give this a shot if you don't mind?

scott:
Hey @Brand-Temp this sounds great.  It should be fairly straightforward, let me know if you have any questions!

scott:
Hi @Brand-Temp hows it coming?

PR #1468, opened 2019-10-03
Add honeypot to new ticket form

schmidt:
Background:

Helpy gives the option to add a recaptcha to the new ticket form (topics/new). If no re-captcha is configured, we should still add some kind of honeypot anti-spam field to the new ticket form. Obviously this could be hidden if the re-captcha is displayed.

Implementation:

This commit adds a URL input to the new topic form, which will hopefully attract spammers to fill in their website. The form is hidden from regular users using `display: none`. In the controller we're then checking if the input was filled. A regular user wouldn't see the input, i.e. it should remain empty. A spammer in contrast happily adds some nonsense and we can reject the input.

When detecting a spammer we're doing the same thing as when the captcha wasn't filled properly, i.e. rendering the form again.

Potential problems:

- I am not sure, if this "feature" is a problem accessibility wise. I simply don't know if screen readers would also "show" the honeypot or not.
- I've seen this approach work reliably on other websites. However, an attacker, who's directly targetting helpy will of course be able to work around the honey pot quite easily.

### DIFF
--- a/app/assets/stylesheets/shared.scss
+++ b/app/assets/stylesheets/shared.scss
@@ -61,3 +61,7 @@
     top: 0;
   }
 }
+
+#topic-url {
+  display: none;
+}
--- a/app/controllers/topics_controller.rb
+++ b/app/controllers/topics_controller.rb
@@ -114,6 +114,11 @@ class TopicsController < ApplicationController
 
     associate_with_doc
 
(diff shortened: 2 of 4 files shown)


schmidt:
This fixes #961 and is a contribution in the context of hacktoberfest.

PR #1469, opened 2019-10-03
Fix 404 after failed submitting new topic with failed captcha

schmidt:
Background:

Anonymous users may enter new topics. Per configuration, this feature is protected by a captcha. When the user fails the captcha test, we want to show the input form again, so that they may re-enter their data and the captcha

Problem: 

Currently helpy renders a 404 due to a missing template for `topics/create` when the captcha verification fails. This is happening due to operator precedence in combination with optional parenthesis in ruby. The code

```
render :new && return
```

is interpreted as

```
render(:new && return)
```

I.e. it's returning before `render` is called. Therefore rails defaults to rendering the current action, which is `create` and there's no `create` template. Consequently Rails renders a 404.

Solution:

There are two common solutions to this:

```
render(:new) && return
```

and

```
return render :new
```

Both version ensure, that the render call is executed before the `return`. I've chosen the latter, because this adheres to the preference in the controller of not using parenthesis. 

This change also adds a test, that exposes the bug.


test/controllers/topics_controller_test.rb
```diff
@@ -273,14 +273,41 @@ class TopicsControllerTest < ActionController::TestCase
     end
   end
 
+  test 'a browsing user should not be able to create a new topic when failing the recaptcha test' do
+    # Make sure recaptcha is set up
+    Recaptcha.configuration.skip_verify_env.delete("test")
```
schmidt: This seems to be the only way to enable the recaptcha verification in the test env. See https://github.com/ambethia/recaptcha#testing

test/controllers/topics_controller_test.rb
```diff
@@ -273,14 +273,41 @@ class TopicsControllerTest < ActionController::TestCase
...
+    assert_select "#new_topic"
+
+  ensure
+    Recaptcha.configuration.skip_verify_env << "test"
+  end
+
   test 'a browsing user should be able to create a new public topic without signing in when recaptcha enable' do
 
     # Make sure recaptcha site_key is set
+    AppSettings['settings.recaptcha_enabled'] = "1"
```
schmidt: Without this `1`, the code wouldn't even try to verify the captcha.

test/controllers/topics_controller_test.rb
```diff
@@ -273,14 +273,41 @@ class TopicsControllerTest < ActionController::TestCase
...
     AppSettings['settings.recaptcha_site_key'] = "some-key"
     AppSettings['settings.recaptcha_api_key'] = "some-key"
 
-    #TopicsController.expects(:verify_recaptcha).returns(true)
```
schmidt: Removed the unused mock, since mocking `verify_recaptcha` isn't even necessary. As documented in the recaptcha readme, verification is disabled in test-mode anyways.

schmidt:
This bug was discovered while working on #961 / #1468 and it would be great if it was considered to be part of hacktoberfest.

schmidt:
Updated the PR to fix invalid block rescue syntax for Ruby 2.4.

PR #1470, opened 2019-10-03
Add tags API to create and manage tags

schmidt:
Background:

Admins may manage tags using the admin UI, but they are not managable via API.

This commit adds a tags API to read, create, and update tags for admin users.

I did not implement deleting of tags, since many of the other APIs also don't implement deletion and I thought, that there might be a good reason for that.

The tags API is only available to admins, similarly to the tags admin user interface.

I'm only exposing the basic attributes of tags, that are also available in the user interface.

### DIFF
--- a/test/controllers/api/v1/tags_test.rb
+++ b/test/controllers/api/v1/tags_test.rb
@@ -1,0 +1,103 @@
+# == Schema Information
+#
+# Table name: tags
+#
+#  id                 :integer          not null, primary key
+#  name               :string
+#  description        :text
+#  color              :string
+#
+
+require 'test_helper'
+
+class API::V1::TagsTest < ActiveSupport::TestCase
+  include Rack::Test::Methods
+
+  def app
+    Rails.application
+  end
+
+  setup do
+    set_default_settings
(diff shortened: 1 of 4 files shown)


schmidt:
This fixes #1202 and is a contribution in the context of hacktoberfest.

test/controllers/api/v1/tags_test.rb
```diff
@@ -0,0 +1,103 @@
...
+  end
+
+  setup do
+    set_default_settings
+    @user = users(:admin)
+    @api_key = ApiKey.create(name: "Test Tags Key", user: @user)
+    @default_params = { token: @api_key.access_token }
+
+    @simple_tag = ActsAsTaggableOn::Tag.new(name: 'simple', description: nil, color: nil)
+    @full_tag = ActsAsTaggableOn::Tag.new(name: 'full', description: 'description', color: '#336699')
```
schmidt: During my tests, there were no tags in the db. If you wish, I can try to move this setup into fixture files.

scott: I think this is fine

PR #1667, opened 2020-03-27
update Persian translation

hadifarnoud:


hadifarnoud:
this pull request contains some RTL fixes as well

hadifarnoud:
Emails used to be fine with RTL but the support is broken now.

it looks like the ticket url in email is also incorrect (notice the url has `en` in it but the only locale available is `fa` and it's set as default too)
Screen Shot 2020-05-21 at 01 02 58

I'm struggling to fix the email issue as you can see from my comments

hadifarnoud:
@scott any idea when this will merge?

hadifarnoud:
we have rewrote the entire way RTL works in Helpy. I hope this improves things a lot more

PR #2084, opened 2024-11-18
Add devcontainer

rip747:
Create a devcontainer to make development and
contribution easier.

The devcontainer is based on the Docker production image with the exception that it uses puma as
instead of nginx.

I needed to pin some of the gems that were being
used in development as they were causing issues.
Those gems were

byebug
pry-byebug
rubocop

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -10,3 +10,10 @@ RUN apt-get update \
   && apt-get install -y nodejs postgresql-client imagemagick --no-install-recommends \
   && apt-get clean \
   && rm -rf /var/lib/apt/lists/*
+
+# chrome needed for running test
+# https://stackoverflow.com/a/71008100/16682649
+RUN apt-get install -y wget
+RUN wget -q -O - https://dl-ssl.google.com/linux/linux_signing_key.pub | apt-key add - \ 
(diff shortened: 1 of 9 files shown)


rip747:
FYI... I'm going to be working on trying to get the system test passing and then I'll rebase for a final. I just keep getting road blocked by:

Error:
BrowsingUserTicketFlowsTest#test_a_browsing_user_should_be_able_to_create_a_private_ticket_via_widget:
Webdrivers::NetworkError: Net::HTTPServerException: 404 "Not Found" with https://chromedriver.storage.googleapis.com/131.0.6778.69/chromedriver_linux64.zip
    test/integration/browsing_user_ticket_flows_test.rb:138:in `block in <class:BrowsingUserTicketFlowsTest>'


Seems that the latest chrome isn't working. I installed it using the following in the Dockerfile:

RUN apt-get install -y wget
RUN wget -q -O - https://dl-ssl.google.com/linux/linux_signing_key.pub | apt-key add - \ 
    && echo "deb http://dl.google.com/linux/chrome/deb/ stable main" >> /etc/apt/sources.list.d/google.list
RUN apt-get update && apt-get -y install google-chrome-stable

I will try to see if I can install an earlier version and maybe that will work. If anyone reads this and knows how to do that, please reply. I just getting started with Docker and getting back into rails so I'm very rusty at this point :)

rip747:
@scott This pull request now installs chrome 114 in order to run test and also gets all test passing.

